PR #6, opened 2014-04-17
Add cwd for git commands

dgreif:
If you run this utility in any directory other than the directory containing the target git repository you cannot get the data from the target git repository.  By adding a cwd option, the user can now specify which directory to target when gathering tag/commit information.

I also changed the 'latestTag' functionality to work with msysgit.


stevemao:
fixed in 69cfb5c61bd3c7a6157a980cf8445e8a0b7a902f. Thanks.

PR #18, opened 2014-06-30
Changes for non-github projects like bitbucket

bennekrouf:
Hi

I've been constrained to make some changes to be able to generate log working with bitbucket for managing 2 differences:
- the commit are set in **/commits/** in bitbucket instead of **/commit/**
- issues are logged in a specific JIRA URL (in my case)

I've also added a "perf" section, and it looks like the "refactor" section was not generated. 

I hope my code respects the existing code. If no do not hesitate to ask me for changes, to iterate till the changes can be integrated.

Thanks

### DIFF
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -32,8 +32,11 @@ Graphite is a much more available resource than lead, so we use it to lower the 
 fix(pen): use blue ink instead of red ink
 
 ```
+```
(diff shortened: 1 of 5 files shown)


stevemao:
Closing for a more generic solution #22

stevemao:
Please checkout the rewrite on the master branch and see if the problem still exists. Thanks.

Issue #12, opened 2014-05-17
Expose the `grep` option, add "Other fixes" section


vladikoff:
How about exposing the `grep`([ref](https://github.com/ajoslin/conventional-changelog/blob/master/lib/git.js#L41)) option, this way you can set which conventions you want in the changelog. The additional commits with conventions that don't match Fix, Feature, Breaking would be grouped as  "Other fixes" in the changelog. 

Thoughts?

stevemao:
Please checkout the rewrite on the master branch and see if the problem still exists. Thanks.

PR #20, opened 2014-07-25
Option to include chores in the changelog output

raszi:
I am using your node package (mainly the writer) in my [github changelog](https://www.npmjs.org/package/github-conventional-changelog)
and we want to include the chores in the changelog output.

With this small addition the original behavior won't change but we give the option to include
the chore changes as well.


stevemao:
Closing for a more generic solution #12

vvo:
So is it now feasible from the command line to force chores to be chosen as a starting point for a new version? thx

stevemao:
Yes, you can always overwrite the transform. Read the docs.

PR #27, opened 2014-10-07
fix(git): support msysgit by using --abbrev flag

### LINKED ISSUE #25: function latestTag does not work under Windows shell
hi, if you execute the command inside function `latestTag` under Windows shell (not the Git shell) throws an error `unknow option max-count=1` due to backtick.

I have made a quick test by making two calls instead of only one, using the output of a command to pass to the other one. With this change, it finds the tag.

dlukez:
As discussed in #25. I had a quick Google and didn't find any negative consequences of using this command, but I've only tested this on Windows with msysgit.

Closes #25

### DIFF
--- a/lib/git.js
+++ b/lib/git.js
@@ -11,7 +11,7 @@ module.exports = {
 //Get latest tag, or if no tag first commit
 function latestTag(done) {
   //Get tags sorted by date
-  cp.exec("git describe --tags `git rev-list --tags --max-count=1`", function(err, stdout, stderr) {
+  cp.exec("git describe --tags --abbrev=0", function(err, stdout, stderr) {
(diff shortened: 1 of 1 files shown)


felixzapata:
this pull request has not been accepted yet?

ajoslin:
I think this is good to merge.

stevemao:
Sorry Windows support wasn't on the top of my list.
Just tested this and it works the same on mac. Will merge this soon.

ajoslin:
I couldn't get this commit to merge properly, so I just added it myself via 69cfb5c61bd3c7a6157a980cf8445e8a0b7a902f

Thanks @dlukez for the suggestion.

Issue #22, opened 2014-08-15
Split up this repo into commits parsing and writing


pkozlowski-opensource:
@ajoslin I'm working on several build-related topics atm and can see number of use-cases where I could use just the git-related part, ex.:
* figure out next semver based on the commit names
* use completely different writing format / target (I think GH releases were mentioned).

In short: IMO splitting up the repo into 2 distinct parts would make each part more useful and easier to maintain. The interface between 2 parts would be output generated by git parsing. 

I can surely just require `lib/git.js` from this repo (might break without warning) or just fork the whole thing (not so great for the community / ecosystem)  but yeh, I feel like the sum of parts could be greater than the combined thing. WDYT? I know you are busy on other topics but I will need this type of think so willing to put some effort into it.

brunobatista:
+1

stevemao:
I'm working on this issue. I feel it's very hard to maintain/test now if they are not broken into smaller pieces.

pkozlowski-opensource:
Great @stevemao !

stevemao:
Thanks @pkozlowski-opensource , I have done https://github.com/stevemao/git-raw-commits and https://github.com/stevemao/conventional-commits-parser and I'm trying to do a writer. Those modules are pretty much based on this one but I added some other features. Please make comments/suggestions on these modules. 

I have problems making the writer. I want the templating system of the writer as flexible as possible. Should I expose a `lodash.template` string? And because in https://github.com/stevemao/conventional-commits-parser I can get any part of a commit so this way it will be very flexible but a `lodash.template` string is pretty hard to construct especially one with `if` statements and nested loops, etc (I have problems making a default one but I think I can make it eventually). Or we can expose something like 

```js
{
  headerTemplate: '...',
  closingIssueTemplate: '...',
  bodyTemplate: '...',
  footerTemplate: '...'
}
```

This way it is very similar to the current implementation and just expose the template. But things like loops cannot be defined by users.

I think `lodash.template` is hard to use especially if there are too many `if` statements or nested loops. May be I should use something else?

ajoslin:
@stevemao maybe it would be best to use a [nunjucks](http://mozilla.github.io/nunjucks/) template or similar to render the whole changelog... example:

```
{% block header %}
{{"###" if bump.patch else "##"}} {{title}} {{subtitle}} {{version}}
{% endblock %}

{% block body %}
{% for section in sections %}
#### {{section.name}}

{% for subject in section.subjects %}
[...]

PR #29, opened 2014-10-22
feat(generate): add function for getting a recommended version bump

hugowetterberg:
The conventional changelog provides an excellent guidance on which part of the version number one should bump on release. This commit exposes a `bump` function that looks at the commits and returns a recommendation for either a `major`, `minor`, or `patch` bump.

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,6 +5,7 @@ var writer = require('./lib/writer');
 var extend = require('lodash.assign');
 
 module.exports = generate;
+module.exports.bump = bump;
 
 function generate(options, done) {
   options = extend({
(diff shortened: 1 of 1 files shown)


stevemao:
I think it's very interesting. However I think it might be better to make this as a standalone module.

voxpelli:
Either the recommended version bump needs to be part of conventional-changelog or conventional-changelog needs to expose the parsed changes to an external module so that it can make a decision based on it before it sends a version number into the standard conventional-changelog method to have the changelog itself written.

stevemao:
@voxpelli I'm working on this now #22

stevemao:
Please checkout https://github.com/stevemao/conventional-recommended-bump

PR #31, opened 2014-10-29
feat(cli): Add command line interface

mikberg:
Add simple command line interface, which can be installed and used globally,
called "changelog". The following is the CLI's help text:

```
  Generate a changelog from git metadata, using the AngularJS commit
  conventions.

  Specify a version as first argument, or use 'next' as version number if no
  version is specified.

  Usage
    changelog
    changelog <version>
    changelog <version> --file <filename>
    changelog <version> --file <filename> --no-write

  Options
    --subtitle    A string to display after the version title in the changelog.
                  For example, it will show "## 1.0.0 "Super Version" if
                  codename "Super Version" is given.
    --repository  If this is provided, allows issues and commit hashes to be
                  linked to the actual commit. usually used with github
                  repositories. E.g. http://github.com/joyent/node
    --from        Which commit the changelog should start at. By default, uses
                  previous tag, or if no previous tag the first commit.
    --to          Which commit the changelog should end at. By default, uses
                  HEAD.
    --file        A file to read the current changelog from and prepend the new
                  changelog's contents to.
    --no-write    Don't write back to the file. Used in combination with --file.
```

If run with no arguments or options, the tool outputs the changelog since last
tag with version number `next`, practical for checking the "current" changelog.
Other options are like the API, except `no-write`, which tells the tool to not
write the new log entires to the file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,12 +20,12 @@ Adapted from code originally written by @vojtajina and @btford in [grunt-convent
 ## Roadmap
 
 - Make it return a stream
-- Add a proper command line interface
+- ~~Add a proper command line interface~~
 - Add configurable subjects & sections
 
 ## Documentation
 
-Simple usage: 
+Simple usage:
 
 ```js
 require('conventional-changelog')({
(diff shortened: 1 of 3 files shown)


btford:
:+1:

stevemao:
Because of #22 the public API will be changed but I will add cli support based on this one. 

@btford npm permissions for btford/grunt-conventional-changelog?

https://github.com/btford/grunt-conventional-changelog/issues/52#issuecomment-82791516

PR #36, opened 2015-01-30
feat(changelog): add currentVersion, bump options

mgartner:
Allows the currentVersion and bump options to be used in place of the version option. Using these options will automatically increment the version number written to the changelog.

The motivation for this is that conventional-changelog is hard to use with `npm version major/minor/patch` commands. These commands automatically increment the `package.json` version number, make a new commit on the branch, and tag the commit with the latest version. However, you can't run `npm version major/minor/patch` before updating the changelog, because conventional-changelog will parse no commits since a new tag has just been created, and you can't run `npm version major/minor/patch` after conventional-changelog because the version written to the changelog will be the old version.

This feature allows a much easier flow for using this tool with the `npm version` command by allowing the user to pass in the `currentVersion` and `bump` to automatically write a new version to the changelog. The workflow then becomes and easy two-step process:

1. Update the changelog with conventional-changelog passing the current version in `package.json` and `major`, `minor`, or `patch`.
2. Run `npm version major/minor/patch` using the same bump level used when making the changelog.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,10 @@ By default, calls the callback with a string containing a changelog from the pre
 
 * `warn` `{function()}` - What warn function to use. For example, `{warn: grunt.log.writeln}`. By default, uses `console.warn`.
 
+* `currentVersion` `{string}` - Used with `bump` in place of `version` to automatically increment the version number written to the changelog. Should be a string in the form: `1.2.3`.
+
+* `bump` `{string}` - Used with `currentVersion to automatically increment the version number written to the changelog. Passing `major` increments the major version number, `minor` increments the minor number, and `patch` increments the patch number.
+
 ## License
 
 BSD
--- a/index.js
+++ b/index.js
@@ -3,6 +3,7 @@ var fs = require('fs');
 var git = require('./lib/git');
 var writer = require('./lib/writer');
 var extend = require('lodash.assign');
+var versioner = require('./lib/versioner');
 
 module.exports = generate;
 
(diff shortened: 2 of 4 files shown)


stevemao:
Version bump might be out of the scope of this module and https://github.com/geddski/grunt-release is doing this. As mentioned in https://github.com/btford/grunt-conventional-changelog/issues/46 the usual workflow is:

* Make changes
* Commit those changes
* Bump
* grunt changelog
* Commit
* Tag

So the bump is done before `changelog` and changelog just need to read the version number from package.json. Closing this for now and if you have any other thoughts just reply and we can reopen. Thanks.

stevemao:
To your point, no we can't use npm version major/minor/patch command because it auto creates a version commit and tag, which is not the usual workflow

mgartner:
Just curious why you say using `npm version` is not the "usual workflow". Have you polled users of this package, or have any source that suggests not to use `npm version` in general? It seems to be a useful tool that saves a manual steps just to bump a version number and tag a commit.

stevemao:
Sorry I didn't make it clear.

`npm version` will also create a version commit and tag. However most of the angular related projects would probably run `changelog` between bump and commit so `npm version` may not be suitable for this workflow. Checkout https://github.com/yeoman/generator-angular/blob/master/Gruntfile.js#L88 and https://github.com/karma-runner/karma/blob/master/Gruntfile.coffee#L161 or other angular related projects as examples.

I'm not saying version bump is not useful, its [separation of concerns](https://github.com/substack/browserify-handbook#module-philosophy). This module should **just focus on generating a changelog**, there are other modules that can do version bump (https://github.com/geddski/grunt-release, https://github.com/vojtajina/grunt-bump, https://github.com/gruntjs/grunt-contrib-bump and [many others](https://www.npmjs.com/search?q=bump)). More references: http://en.wikipedia.org/wiki/Separation_of_concerns

mgartner:
I understand your concern, but why neglect one workflow just because it's not what you use? If a user prefers to use `npm version`, then it is impossible to use this package now that you've disallowed a user from providing a custom version. I've also made these additional settings optional so a user can choose which workflow works better for them.

stevemao:
Again, you can use any tool with this module but remember, *This module should just focus on generating a changelog*. It is possible to provide a custom version. https://github.com/ajoslin/conventional-changelog/blob/master/lib/writer.js#L84

PR #38, opened 2015-02-13
feat(defaults): version and repository are read from package.json

stevemao:
as discussed in btford/grunt-conventional-changelog#44, we think it is better if we move more logic to this main module

these defaults would also be great for cli

### DIFF
--- a/lib/writer.js
+++ b/lib/writer.js
@@ -1,7 +1,11 @@
 var dateFormat = require('dateformat');
 var es = require('event-stream');
-var util = require('util');
 var extend = require('lodash.assign');
+var fs = require('fs');
+var normalizeData = require('normalize-package-data');
+var parseGithubUrl = require('github-url-from-git');
+var url = require('url');
+var util = require('util');
 
 var VERSION = '## %s%s';
 var PATCH_VERSION = '### %s%s';
@@ -11,6 +15,20 @@ var LINK_COMMIT = '[%s](%s/commit/%s)';
 var COMMIT = '(%s)';
 var HEADER_TPL = '<a name="%s"></a>\n%s (%s)\n\n';
 var EMPTY_COMPONENT = '$$';
+
+function parseNonGithubUrl(nonGithubUrl) {
+  try {
+    var idx = nonGithubUrl.indexOf('@');
+    if (idx !== -1) {
+      nonGithubUrl = nonGithubUrl.slice(idx + 1).replace(/:([^\d]+)/, '/$1');
+    }
+    nonGithubUrl = url.parse(nonGithubUrl);
+    var protocol = nonGithubUrl.protocol === 'https:' ?
+      'https:' : 'http:';
+    return protocol + '//' + (nonGithubUrl.host || '') +
+      nonGithubUrl.path.replace(/\.git$/, '');
+  } catch (e) {}
+}
 
 function getVersion(version, subtitle) {
   subtitle = subtitle ? ' ' + subtitle : '';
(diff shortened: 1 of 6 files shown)


stevemao:
Hi @ajoslin , if its OK with you I'm happy to help to manage this repo whenever I'm free. This one looks quite interesting but not too busy so I have confidence. Thanks.

ajoslin:
Just gave you push access. Thanks a lot @stevemao!

This repository is so behind on maintenance, every time I get an email about someone opening a PR I get a bit frustrated because I don't have time to spend on it.

Thanks for the offer.

ajoslin:
Feel free to go crazy - and send me an email (email in profile) if you have questions.

stevemao:
Thanks @ajoslin 

@sindresorhus, @eddiemonge and @passy might also be keen to manage.

stevemao:
Currently the coding style might be a bit messy. Once I fix this I'll merge it.

lib/writer.js
```diff
@@ -1,6 +1,10 @@
 var es = require('event-stream');
 var util = require('util');
 var extend = require('lodash.assign');
+var fs = require('fs');
+var normalizeData = require('normalize-package-data');
+var parseGitHubURL = require("github-url-from-git");
+var url_ = require("url");
```
eddiemonge: would it be horrible to alphabetize this list? and why `url_` instead of just `url`?

lib/writer.js
```diff
@@ -147,3 +162,44 @@ function currentDate() {
...
+    var protocol = url.protocol === "https:"
+      ? "https:"
+      : "http:"
+    return protocol + "//" + (url.host || "") +
+      url.path.replace(/\.git$/, "");
+  }
+  catch(e) {}
+}
+
+function getPackageData(pkg) {
```
eddiemonge: feel like these should be near the top with the other get functions

lib/writer.js
stevemao: I copied this from npm code. There is a local url below.

lib/writer.js
```diff
@@ -147,3 +162,44 @@ function currentDate() {
...
+  var pkgData = JSON.parse(fs.readFileSync(pkg, {encoding:'utf8'})) || {};
```
eddiemonge: why not:

```
var pkgData = {};
try { pkgData = require(pkg); } catch (e) {}
```
other than it being one more line?

lib/writer.js
```diff
@@ -147,3 +162,44 @@ function currentDate() {
...
+  normalizeData(pkgData);
```
eddiemonge: so wish this returned the normalized data instead of mutating the original

lib/writer.js
```diff
@@ -147,3 +162,44 @@ function currentDate() {
...
+
+  url = (url && url.indexOf("github") > -1)
+    ? parseGitHubURL(url)
+    : nonGithubUrl(url);
+
+  if (!url) {
+    return null;
+  }
+
+  return url;
```
eddiemonge: `return url || null;`

lib/writer.js
eddiemonge: shouldnt the local have a different name so you know `url` is from the require?

stevemao: true. `npm`'s coding style is weird. Sorry for being a no brainer but just copy other people's code.

lib/writer.js
```diff
@@ -147,3 +162,44 @@ function currentDate() {
 
   return util.format('%d-%s-%s', now.getFullYear(), pad(now.getMonth() + 1), pad(now.getDate()));
 }
+
+function nonGithubUrl(url) {
```
ajoslin: Why the try catch block?

lib/writer.js
ajoslin: We can't use require() because it will require relative to the conventional-changleog library, not the cwd.

lib/writer.js
ajoslin: Or better yet, just `return url` and the falsy value will just be a falsy value.

lib/writer.js
stevemao: I was wondering the same thing when I copied... Will fix this later :)

lib/writer.js
stevemao: True, I thought it would. `npm` is very weird.

lib/writer.js
sindresorhus: Pretty sure https://github.com/npm/normalize-package-data#what-normalization-currently-entails normalizes this already.

stevemao: It doesn't actually fix non github url. https://github.com/npm/npm/blob/master/lib/repo.js#L69 fixes it.

PR #39, opened 2015-02-17
docs(README): add info about running tests

marcin-wosinek:
I didn't work with mocha before, and I found that information missing.


stevemao:
I think it might be better to make mocha as a dev dep

marcin-wosinek:
I belief doc part is still relevant - without mocha installed globally & available from PATH we are still getting errors

stevemao:
This should be fixed. Can you try the master branch? If it still occurs let us know. Thanks.

marcin-wosinek:
ah my bad. after checkout & reinstall it works as expected. Great, thanks:)

PR #44, opened 2015-03-13
fix(writeLog): fix require statement for Writer

adam-beck:


stevemao:
Thank you. :+1:

stevemao:
Hi @ajoslin , Can you enable travis on https://travis-ci.org? Before this was merged it passed tests on mac but would fail on linux. So this would have been fixed before we release v0.0.13. Thanks.

ajoslin:
Enabled.

stevemao:
Thanks @ajoslin .

Issue #45, opened 2015-03-14
Should be able to easily generate changelog for any previous version(s).


stevemao:

dragn:
+1
The implementation could traverse tags one-by-one and gather changes between them, also making note of the date the version was released (tagged commit date).

stevemao:
Oh that's a good point

stevemao:
http://stackoverflow.com/a/6900369/3304774

Looks like Lightweight Tags do not store date information. So if using a lightweight tag the date would have to be the date of the commit that is tagged on :(.

stevemao:
Also if multiple tags are found, it should probably use the highest semver (filter out the ones that are not semver first).

stevemao:
Please checkout the rewrite on the master branch and see if the problem still exists. Thanks.

PR #51, opened 2015-04-05
feat(split): split into smaller modules

stevemao:
**DO NOT MERGE THIS**

Use git-raw-commits, conventional-commits-parser and conventional-commits-writer.

BREAKING CHANGE:

This module is rewritten so API is changed and it is not backward compatible. Please check conventional-commits-parser and conventional-commits-writer for more information.

@ajoslin , @btford, @vojtajina, @pkozlowski-opensource I know you guys are busy with other stuff but this is very crucial to this tool as it can close almost all the issues here and plus more.

Could you guys please review this and all the related modules? You guys are amazing and I really wan to know your opinions.

Thanks all.


ajoslin:
I'll review this tonight.

stevemao:
Thank you @ajoslin

bendrucker:
If you want a somewhat more robust git log parser check out https://github.com/bendrucker/git-log-parser

bendrucker:
Also the commit parser shouldn't have a stream interface since it's not actually parsing line by line. You can just make it a plain sync function and then use https://github.com/thlorenz/transformify here to wrap it in a stream interface.

stevemao:
@bendrucker How does it work? The parser needs to understand different parts of a conventional commits but your module is not doing it.

bendrucker:
git-log-parser is just returning an object stream of commits with the documented object structure. Then you'd pipe it through something like conventional-commits-parser to actually parse the commit objects that get emitted. The log parsing benefits from streaming through stdin but everything else is actually synchronous no matter what interface you wrap around it.

stevemao:
You get benefits when you want to generate all logs from the past #45 because it one block of a single version each time, meaning you don't need to wait for the whole log. If you just need to generate the latest one you won't get benefits because there are grouping and reordering logics in the writer so it has to wait until it gets all the commits of a version.

I still don't understand how  `git-log-parser` can bring more benefits over the current one. I just need a bunch of raw commits. If it errors I just simply emit an error.

Sorry if I miss anything.

stevemao:
Alright need to rewrite this...

PR #52, opened 2015-05-12
docs(CONVENTIONS): explain how to format breaking changes

huerlisi:
The CONVENTIONS.md doc does mention 'breaking changes', but does
not explain the actual format used.

This patch adds some sentences explaining how to markup your commits
to generate beaking changes sections.

### DIFF
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -46,5 +46,7 @@ The body should include the motivation for the change and contrast this with pre
 The footer should contain any information about **Breaking Changes** and is also the place to
 reference GitHub issues that this commit **Closes**.
 
+**Breaking Changes** are detected as such if the Body contains a line starting with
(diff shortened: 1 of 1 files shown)


stevemao:
Thanks. Although I was thinking maybe we should remove this file but give a link to angular's commit message guide. Also in the future this would work with other conventional commit messages.

huerlisi:
@stevemao thanks for merging!

Regarding switching this file with a link to the angular's commit message guide: My gut feeling says no:-) I'd rather love it the other way round, that this is the actual reference and Angular points here. The reason is simple (though political): when doing a project with e.g. EmberJS, or C++ it feels a bit strange to point to Angular, but it's okay to point to this generic library that established this convention.

I just think that having this in a not-so-much branded repository keeps adoption rate higher.

Just my 2c.

PR #54, opened 2015-05-21
feat(parseRawCommit): allow for lack of `S` or `:`

gdi2290:
we are human after all

https://github.com/angular/angular/issues/2058

updated regexp
https://regex101.com/r/wZ2oK2/1

### DIFF
--- a/lib/git.js
+++ b/lib/git.js
@@ -67,7 +67,7 @@ function parseRawCommit(raw) {
     }
   });
 
-  match = raw.match(/BREAKING CHANGE:\s([\s\S]*)/);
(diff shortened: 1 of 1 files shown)


stevemao:
Hi @gdi2290 Thanks for the PR

However, I'm not planning to look at the current version anymore and I want to roll out the next version as quickly as possible. So can you take a look at #51? Thanks again.

stevemao:
BTW, I think this issue is fixed in the upcoming release but do checkout. 

Ref: https://github.com/stevemao/conventional-commits-parser/issues/3

stevemao:
Please checkout the rewrite on the master branch and see if the problem still exists. Thanks.

Issue #53, opened 2015-05-13
Improve the docs of commit message guideline 


stevemao:
#52

In the up coming new release we can fully customise the commit message format so a strict commit message guideline is not necessary.

stevemao:
Or we can come up with some non-strict commit message guideline

voxpelli:
I think the strict conventions are one of the pro's with this tool – an easy guideline/standard for your team on how to format commit messages. Even if other alternatives may become supported by this tool I would like to keep the option of enforcing the existing `CONVENTIONS.md` and to keep the file as a reference for that enforcing.

stevemao:
You can still config to make it strict. It just gives people more options and more projects can adapt this.

voxpelli:
@stevemao Like JSCS it would be nice to still have the old style like a "preset" or such: http://jscs.info/overview.html#presets

stevemao:
:+1: I was thinking about that too

bachp:
I also think that having a fixed convention is an advantage of conventional changelog.
And I belive the current format is flexible enough to be working for most people. Choice for the sake of choice is not always good.

stevemao:
OK let me rephrase. I'm not saying there is not going to be strict rules. There are but you can define them. Just like jscs or jshint, you can define the rules. They never restrict you from having your own rules. Same to this project, you don't have to use Angular commit message. If you have your own conventions you can still use this one to generate logs for you based on your commit convention.

bachp:
I understand what you mean.
But I think there should add least be a CONVENTIONS.md document describing a best practice standard.
If configuration options are a good or bad thing is another question. But one example I like to mention is go the programming language, where a simple command [go fmt](https://blog.golang.org/go-fmt-your-code) made the whole discussion about coding style a non issue as there is only one style enforced by an automatic tool.
If there is one convention for conventional-changelog people would not need to do things differently for each project they do.

stevemao:
Sorry, CONVENTIONS.md doesn't have to be removed. It probably needs to be renamed and reworded. Point is we need new docs to guide users.

I understand that you like to have consistency across all projects but in real world each project has their own requirements. If you look at other issues in the repo you'll know why I need to make this more flexible.

bachp:
I'm not oposed to make you tool more flexible, it is great and I think to make it useful for more people is a good thing.

So let me rephrase my proposal:
[...]

PR #55, opened 2015-05-21
docs(CONVENTIONS.md): correct regexp

gdi2290:
https://regex101.com/r/zJ8pC7/2


stevemao:
#53

I don't think this is supposed to be a regex (If it was then the newline should be optional). We'll take it into account when writing the new docs. Thanks.

stevemao:
I have pushed the rewrite to the master branch. Would be great if you could have look. Thanks.

PR #60, opened 2015-06-10
feat(non-numeric-issue): Add support for non-numeric issues

### LINKED ISSUE #59: Add support for non-numeric issues
Currently, this plugin only supports numeric issues. When using Jira, issues are (possibly) prefixed with non numeric characters (e.g. PROJECT-123). When adding a comment like 'Closes PROJECT-123', these lines are not appended in the changelog, because it expects a hash, followed by a number.

rsids:
Issues which do not start with a hash followed by a number are now correctly parsed in the changelog

Closes #59

### DIFF
--- a/lib/git.js
+++ b/lib/git.js
@@ -45,21 +45,21 @@ function parseRawCommit(raw) {
   msg.closes = [];
   msg.breaks = [];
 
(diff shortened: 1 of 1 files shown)


stevemao:
Hi @rsids , Can you send a PR against https://github.com/stevemao/conventional-commits-parser/blob/master/lib/regex.js? We are breaking down this module into smaller ones and is waiting on #51. Thanks.

PR #69, opened 2015-07-03
fix(spelling): spelling mistake of host bitbucket

gitawego:


stevemao:
Thanks. Can you add a test for this?

PR #75, opened 2015-07-21
chore(pkgRepo): move version before getPkgRepo

### LINKED ISSUE #74: Version not parsed from package.json if "repository" field not set in file.
The issue occurs on [line 88 of index.js](https://github.com/ajoslin/conventional-changelog/blob/master/index.js#L88). `getPkgRepo` fails if there is no `repository` information, which prevents the version information from being added to the `context` object, causing blank values in the changelog output. 

I'll see if I can get a PR out for this.

Steps to recreate:
1. Have an NPM repository without a `repository` field:
```
{
  "name": "my-package",
  "version": "1.0.1",
}
```
2. Run `conventional-changelog`
3. See that output does not pick up `version` information

klamping:
Closes #74 

See steps to recreate in issue.

### DIFF
--- a/index.js
+++ b/index.js
@@ -85,14 +85,15 @@ function conventinalChangelog(options, context, gitRawCommitsOpts, parserOpts, w
           pkg = pkgObj.value;
           try {
             pkg = JSON.parse(pkg);
-            var repo = getPkgRepo(pkg);
             pkg = options.pkg.transform(pkg);
+            context.version = context.version || pkg.version;
+
+            var repo = getPkgRepo(pkg);
 
             if (repo.type) {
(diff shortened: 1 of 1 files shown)


stevemao:
:+1: It would be great if you can add a test for this?

klamping:
I'll see what I can do. :)

stevemao:
Thanks. You could just add a package.json with only a version.

stevemao:
I'm adding tests now

klamping:
Great, thanks for getting the tests taken care of. I had an event I had to go with so didn't have time.

:+1:

PR #76, opened 2015-07-22
Add all commit types for angular preset

eddywashere:
I noticed there were some commit types missing for the angular preset. Any reason they weren't included? Included a pr just in case.


stevemao:
This is because those ones are ignored by the previous versions. So I'm just doing the same thing. You could checkout the angular repo to see whether these commits are ignored.

stevemao:
Closing now... If you think these types should be included please submit an issue at angular's repo. Thanks.

PR #82, opened 2015-08-10
docs: some thoughts

ajoslin:
The point of this is to spark a discussion about making simpler defaults for this project.

These wording changes assume that we make angular the default and rename it to `standard` or something similar.

Why should there be an opinionated default? Because this project's current defaults just give you a formatted and sorted version of `git log`, which isn't very helpful. 

Why should angular be the default? Because angular's conventions make the most sense and are easy for people to adopt (if we  document them well).

Why rename the angular presets to `standard`? Because lots of people in the npm community hate AngularJS. But whether you like angular or not, the conventions are worth using.

I don't intend this to be merged yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,51 +1,46 @@
 #  [![NPM version][npm-image]][npm-url] [![Build Status][travis-image]][travis-url] [![Dependency Status][daviddm-image]][daviddm-url] [![Coverage Status][coveralls-image]][coveralls-url]
 
-> Generate a changelog from git metadata
+> Generate a changelog from your git commit messages
 
+### Quick Start
 
-## Why
+```
+npm install -g conventional-changelog
 
-- Used by AngularJS and related projects.
-- Everything internally or externally is pluggable.
-- High performant. It doesn't spawn any extra child process to fetch data.
-- Intelligently setup defaults but you can still modify them to your needs.
-- Fully configurable. There are [many presets](presets) that you can use if you just want to use the same conventions. But it is also possible to configure if you want to go down to the nth degree.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -57,29 +52,25 @@ Returns a readable stream.
...
 
 Type: `object`
 
+An object representing package.json.
+
 ###### path
 
 Type: `string` Default: `'package.json'`
 
-The location of your "package.json".
```
ajoslin: There are two options named `transform`. I deleted one, but I'm not really sure what's up there.

README.md
```diff
@@ -57,29 +52,25 @@ Returns a readable stream.
...
-
-###### transform
```
ajoslin: There are two options named transform. I deleted one, but I'm not really sure what's up there.

stevemao: Yeah they do different things (bad name). The one under pkg is to tweak the parsed package.json a bit. It is used by `conventional-github-releaser` where you need to prepend a leading `'v'` in version.

stevemao:
The reason I didn't make angular as default are:
1. I don't want to be opinionated.
2. Like you said some people "hate" angular's conventions (hate is probably not accurate).
3. I checked a lot of other projects' conventions and I think they work nicely too. For example: jQuery's convention is also used widely and there are certain nice things about it.

I don't mind making angular as `standard`. I suggest that we should make it a vote? While this PR addresses other issues too so I'll open another one.

stevemao:
In the docs, I think it's good if we ask user to alias the command. and put the changelog script in package.json.

ajoslin:
Nice. Love the new README. :+1:

stevemao:
Thanks for the suggestion @ajoslin :)

PR #83, opened 2015-08-10
docs(conventions/angular): rearrange sections and fix a few errors

stevemao:
The reason to move examples section up is because it was in the middle of explaining different parts of a commit message.
Also the `perf` example is missing and it was not correct.

@ajoslin would like to review this?


ajoslin:
LGMT

stevemao:
Hi @ajoslin, I also add a revert example. Can you review this again. Thanks.

PR #92, opened 2015-08-20
add a note about how to list available parameters on CLI

oupala:
Some users want to have all important informations in the README file.
Some developers do not want to repeat themselves ([DRY](https://en.wikipedia.org/wiki/Don't_repeat_yourself)).

To balance between these two tendancies, I propose to explain all users how to get easily the list of available parameters on the command line.


stevemao:
Thanks @oupala :)

PR #94, opened 2015-08-25
Fixed linkCompare not working

el-davo:
linkCompare does not seem to work. This is the only way i could get linkCompare to get picked up properly. Unless there is another way?


stevemao:
`linkCompare` is on the `context` object

Checkout the test for example
https://github.com/ajoslin/conventional-changelog/blob/master/test/test.js#L483-L543
I don't know what you are trying to fix. If you think there is a bug give more examples or PR with a fail test.

Issue #98, opened 2015-09-02
If get-pkg-repo fails to get info, it should fall back to the original repo url


rsids:
The changelog only includes issues when the repository url is a known url (eg bitbucket or github). When using a private url, the generator doesn't parse the issuenumbers

stevemao:
The intension was to use `get-pkg-repo` to get repo info and if it fails, it should fall back to the original repo url. I'll have to have a look.

rsids:
When you use an unknown repo, you have to specify the host, repo & commit value in the context object. That will relink the commits, but it won't link the issues.
Even if it would, there should be an extra option for the issue links. In the project we're using conventional changelog, the repo has an entirely different url as the issuetracker.

oupala:
@rsids Could give an example about how to specify host, repo & commit value in the context object.

Thanks.

rsids:
```
context: {
      // context goes here
      host: 'https://git.HOST.nl/projects/ON/repos',
      repository: 'REPONAME',
      issue: 'issues',
      linkReferences: false
},
```
I've just found that it won't link the issues since I hadn't specified the referenceActions in the parserOpts. Which shouldn't be necessary, since I just want to use the defaults.
With the referenceActions specified, the issues are appended, but all prefixed with a #, not with the prefix I specified in issuePrefixes, see #99.
Furthermore, our git repo is at https://git.HOST.nl/..., while are issuetracker is as https://jira.HOST.nl/..., now, all the issues are linkt to git.HOST.nl.

stevemao:
for host defaults, we only support three. PR welcome for more hosts #80

stevemao:
Blocked by #91 which is blocked by https://github.com/sindresorhus/read-pkg/issues/2

stevemao:
Also @rsids please keep this issue focused. I already opened #80 to track for more host support. Let's keep this for the fallback if repo cannot be parsed by `get-pkg-repo`

destroyerofbuilds:
I'm encountering the same issue while working with a package hosted on an on-premise GitLab instance.

Here's the results of running `get-pkg-repo` directly:

```bash
get-pkg-repo package.json
{ protocol: 'https:',
  slashes: true,
  auth: null,
  host: 'gitlab.example.com',
  port: null,
  hostname: 'gitlab.example.com',
  hash: null,
  search: null,
  query: null,
  pathname: '/mygroup/mymodule.git',
  path: '/mygroup/mymodule.git',
[...]

PR #104, opened 2015-09-12
[WIP] Add "private" property to repository object

raido:
This private property is used as repository type when it is not matched from the repo url.

For example:

"repository": {
    "type": "git",
    "url": "https://on.premises.com/ajoslin/conventional-changelog.git",
    "private": "gitlab"
}

I tested this with JSHint preset and it generated correct paths to the commit and issues.

I am not sure if this is the way to go but i thought i would give it a try.

Relates to #98

### DIFF
--- a/index.js
+++ b/index.js
@@ -136,7 +136,7 @@ function conventionalChangelog(options, context, gitRawCommitsOpts, parserOpts, 
           if (match) {
             type = match[0];
           } else {
-            if (pkg.repository.private) {
+            if (pkg && pkg.repository && pkg.repository.private) {
               type = pkg.repository.private;
             }
           }
--- a/index.js
+++ b/index.js
@@ -105,12 +105,12 @@ function conventionalChangelog(options, context, gitRawCommitsOpts, parserOpts, 
 
           repo = getPkgRepo(pkg);
 
-          if (repo.type) {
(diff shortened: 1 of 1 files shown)


stevemao:
The problem should be: If get-pkg-repo fails to get info, it should fall back to the original repo url

raido:
Yes, but it still would not match what type of host is being used. I don't see a way how from a random URL one could figure out if this is a gitlab on premises or not.

stevemao:
There is already `context.host`

raido:
Could you explain a bit more?

stevemao:
https://github.com/stevemao/conventional-changelog-writer#host

raido:
I am coming around how this all works, context file and so on. I can add anything to the context object which will be directly accessable from the templates. This means i can really simply integrate tools like JIRA and so on.

So overall only thing that was missing, was the "context.host" from the if statement, i am going to remove this "type" property i added, because this is not relevant. Everything can be configured via context.

raido:
Actually it seems, there's nothing relevant in my PR. Having parserOpts and context defined correctly, it all works as i'd expect for my use case. Closing this.

stevemao:
What you need to do is PR against https://github.com/ajoslin/conventional-changelog/issues/80

PR #105, opened 2015-09-12
chore(tests): add --template . to git init calls

raido:
This fixes issues with tests, if local system has git templates defined, commit-msg hooks which should confirm that commit messages are valid etc. causing tests to fail.

### DIFF
--- a/test/express.js
+++ b/test/express.js
@@ -10,7 +10,7 @@ describe('presets', function() {
   describe('express', function() {
     before(function(done) {
       shell.cd('express');
-      shell.exec('git init');
+      shell.exec('git init --template=../git-templates');
       writeFileSync('test1', '');
       child.exec('git add --all && git commit -m"deps: type-is@~1.6.3\n\n - deps: mime-types@~2.1.1\n - perf: reduce try block size\n - perf: remove bitwise operations"', function() {
         writeFileSync('test2', '');
--- a/test/angular.js
+++ b/test/angular.js
@@ -10,7 +10,7 @@ describe('presets', function() {
   describe('angular', function() {
     before(function(done) {
       shell.cd('angular');
-      shell.exec('git init');
+      shell.exec('git init --template=../git-templates');
       writeFileSync('test1', '');
       shell.exec('git add --all && git commit -m"chore: first commit"');
       writeFileSync('test2', '');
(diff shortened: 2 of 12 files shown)


stevemao:
I don't understand. Are you saying if there is a commit-msg hook in your project this would fail?

stevemao:
Just IMO I think git-related tests are mainly for CI. Reasons are:
1. Users might not have git installed.
2. Users have different versions of git and different environments setup.

As you can see now CI fails so I can't merge this.

stevemao:
If you think what you have changed reflects to most scenarios, fix the CI test and provide more details on why tests would fail before the changes. Thanks.

raido:
This is a rare case, but given example based on my configuration, then:

* I have ~/.git_template/hooks/commit-msg

This commit-msg does check to third party tool using CURL to confirm that commit message has a correct syntax. Now when i first checked out this project and ran "npm test" i had failing tests, because all my .git_template/hooks also get added to test case "git init" generated repositories. 

I might not be a common case, but it definitely saves some debugging time why tests are failing for other developers who have similar setups using commit message validations.

test/express.js
```diff
@@ -44,7 +44,7 @@ describe('presets', function() {
 
           expect(chunk).to.include('### Dependencies');
           expect(chunk).to.include('type-is@~1.6.3');
-          expect(chunk).to.include(' - deps: mime-types@~2.1.1\n');
+          expect(chunk).to.include('- deps: mime-types@~2.1.1\n');
```
stevemao: This change is not necessary.

stevemao:
Looks good to me otherwise :)

raido:
Well, Travis build now fails. Looking at logs "...type-is@~1.6.3\n\n- deps: mime-types@~2.1.1".. cannot match " - deps: mime-types@~2.1.1" - because there is a whitespace missing. On my local setup all tests pass but on Travis somehow whitespace goes missing.

raido:
I did checkout based on Travis log and it now fails locally too.

raido:
I think i have narrowed it down why express test fails. I downgraded parser from v0.1.0 to v.0.0.19 - then express test passed, but ember test begin to fail.  This clearly confirms that something has changed within the parser.

test/express.js
stevemao: This should be fixed now. Can you please remove this change? I'll merge this as soon as I get a Mac. Thanks.

raido:
lgtm

stevemao:
can you squash the commits?

raido:
Squashed.

stevemao:
Merged. Thanks @raido

PR #106, opened 2015-09-13
Basic JIRA preset

raido:
Probably needs more tweaking and i am happy to implement any changes. Current implementation depends on context.jira_browse property, which should be a full path to https://jira.instance.com/browse. This is used in the templates to generate links like: https://jira.instance.com/browse/PROJECT-123


raido:
Travis build is failing because this PR depends on https://github.com/stevemao/conventional-commits-parser/pull/19

stevemao:
JIRA should be a host not a preset?

raido:
This is a preset because, one can have self hosted JIRA environment. I don't see how it could be host. If it can, please elaborate.

If understand correctly, we should be able to define a type of host which then would checkup "host" configuration from "hosts/", otherwise i cannot see how it could automatically match it.

stevemao:
A preset is usually for a well defined commit message. Eg angular follows a very strict message format. But it can be hosted on different hosts. Eg GitHub and bitbucket.

raido:
Sorry for blowing this one up. I was not paying attention in which repo i were.

Issue #100, opened 2015-09-03
Cli flag name confusing


oupala:
The module is offering an [append option](https://github.com/ajoslin/conventional-changelog#append).

For reverse CHANGELOGs (blog style), it could be useful to offer a **prepend option**.

stevemao:
The log is either in append or prepend mode (the current behaviour)

oupala:
You mean that prepend mode is the default mode.

Append mode can be choosen by the `-a` option.

Right?

stevemao:
Yeah

oupala:
Ok, I tested it, prepending is possible.

It is however a bit complicated to have this behaviour. As far as I tested, we are obliged to specify the prepended file twice!

```bash
conventional-changelog --preset angular --context changelog.json --infile CHANGELOG.md --outfile CHANGELOG.md --release-count 1
```

oupala:
In fact, the overwrite option allows to prepend the file:

```bash
-w, --overwrite           Overwrite the infile
```

The resulting command line is then:

```bash
conventional-changelog --preset angular --context changelog.json --infile CHANGELOG.md --overwrite CHANGELOG.md --release-count 1
```

But using an **overwrite** parameter for **prepending** is really confusing as it is not overwriting at all.

Maybe the parameter name should be changed as the following:

```bash
-s, --same           Ouputting to the infile in prepending mode
```

In fact, the most confusing thing is that it is the `--release-count` that set if the file will be overwritten or not.

- `--release-count 1` => file prepended
- `--release-count 0` => file overwritten

stevemao:
I agree the flag names might be a bit confusing

stevemao:
Also we should probably remove `--raw-commits-opts`, `--parser-opts` and `--writer-opts` and expose useful options of them with each single flag. `--context` can still be a json file.

stevemao:
@oupala you happy with the cli help message?

https://github.com/stevemao/conventional-changelog-cli/blob/master/cli.js#L13-L46

oupala:
@stevemao Yes, it's good. Except that you kept the spelling mistake I made: change `Ouputting` to `Outputting`.

You also use camelCase sometime while you're using lowercase sometime. For example `sameFile` and `infile` and `outfile` (see https://github.com/stevemao/conventional-changelog-cli/blob/master/cli.js#L50-L52).

stevemao:
Good catch @oupala !

[...]

Issue #80, opened 2015-08-05
More hosts


stevemao:
~~Some people mentioned JIRA~~

CC @rsids @bennekrouf @z0r1k 

bitbucket is partially supported apparently.

rsids:
JIRA is selfhosted, so it will be hard to support it out of the box

stevemao:
cc @Rauno56

destroyerofbuilds:
On-premise GitLab.

stevemao:
Just an update: JIRA is not a host and it is supported now.
Any other hosts are welcome.

admosity:
Saw #87 linking to here. Experiencing it with standard-changelog 1.0.1 

Am I missing some steps/can someone point me to where I should add this implementation?

Thanks!

stevemao:
@etremblay see #185

admosity:
@stevemao thanks, will check it out when i get a chance

admosity:
@stevemao PR submitted!

hbetts:
Closing out. Looks like there's work to improve how we manage the templates in pull request #187.

However, this issue is really generic. If someone has a specific host they would like added, or there is a host we can support _better_, please feel free to open a new issue.

PR #107, opened 2015-09-17
feat(preset): A gitlab merge request version of the angular preset

wzrdtales:
This adds basically a copy of the angular preset, but it requires the commit to be a merge request in the gitlab style.

### DIFF
--- a/presets/gitlab-angular.js
+++ b/presets/gitlab-angular.js
@@ -1,0 +1,87 @@
+'use strict';
+var compareFunc = require('compare-func');
+var Q = require('q');
+var readFile = Q.denodeify(require('fs').readFile);
+var resolve = require('path').resolve;
+var semver = require('semver');
+var _ = require('lodash');
+
+function presetOpts(cb) {
+  var parserOpts = {
+    headerPattern: /^(\w*)(?:\((.*)\))?\: (.*)$/m,
+    headerCorrespondence: [
+      'type',
+      'scope',
+      'subject'
+    ],
+    noteKeywords: 'BREAKING CHANGE',
+    revertPattern: /^revert:\s([\s\S]*?)\s*This reverts commit (\w*)\./,
+    revertCorrespondence: ['header', 'hash']
+  };
+
+  var writerOpts = {
+    transform: function(commit) {
+      if (commit.type === 'feat') {
+        commit.type = 'Features';
+      } else if (commit.type === 'fix') {
+        commit.type = 'Bug Fixes';
+      } else if (commit.type === 'perf') {
+        commit.type = 'Performance Improvements';
+      } else if (commit.type === 'revert') {
+        commit.type = 'Reverts';
+      } else {
+        return;
+      }
+
+      if (commit.scope === '*') {
+        commit.scope = '';
+      }
+
+      if (typeof commit.hash === 'string') {
+        commit.hash = commit.hash.substring(0, 7);
+      }
+
+      if (typeof commit.subject === 'string') {
+        commit.subject = commit.subject.substring(0, 80);
+      }
+
+      _.map(commit.notes, function(note) {
+        if (note.title === 'BREAKING CHANGE') {
+          note.title = 'BREAKING CHANGES';
+        }
+
+        return note;
+      });
+
+      return commit;
+    },
+    groupBy: 'type',
+    commitGroupsSort: 'title',
+    commitsSort: ['scope', 'subject'],
+    noteGroupsSort: 'title',
+    notesSort: compareFunc,
+    generateOn: function(commit) {
+      return semver.valid(commit.version);
+    }
+  };
+
+  Q.all([
+    readFile(resolve(__dirname, '../templates/angular/template.hbs'), 'utf-8'),
+    readFile(resolve(__dirname, '../templates/angular/header.hbs'), 'utf-8'),
+    readFile(resolve(__dirname, '../templates/angular/commit.hbs'), 'utf-8'),
+    readFile(resolve(__dirname, '../templates/angular/footer.hbs'), 'utf-8')
+  ])
+    .spread(function(template, header, commit, footer) {
+      writerOpts.mainTemplate = template;
+      writerOpts.headerPartial = header;
+      writerOpts.commitPartial = commit;
+      writerOpts.footerPartial = footer;
+
+      cb(null, {
+        parserOpts: parserOpts,
+        writerOpts: writerOpts
+      });
+    });
+}
+
+module.exports = presetOpts;
--- a/presets/gitlab-angular.js
+++ b/presets/gitlab-angular.js
@@ -6,14 +6,17 @@ var resolve = require('path').resolve;
 var semver = require('semver');
 var _ = require('lodash');
 
+var bodyPattern = /^(\w*)(?:\((.*)\))?\: (.*)$/m;
+var headerCorrespondence = [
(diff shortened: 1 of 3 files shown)


wzrdtales:
@ajoslin  I don't know yet why the express test is failing, I have not touched it though.

stevemao:
what does this do exactly?

wzrdtales:
@stevemao Like already said it is the angular preset, but it filters only for merge requests and takes the first entry found instead of the subject line, as the subject line is naturally the merge message.

stevemao:
The angular project doesn't use any merge commits to reduce noise. Also merge commits has nothing to do with your git host?

wzrdtales:
Considering Gitlab, the only special case here is the format how gitlab displays merge request in the commits and every feature, bugfix e.g. are in our case always a merge request. We want to use your module to auto generate the changelog in the way of the angular conventions, but we still want to keep the merge messages as the title of the merge request. You can change the subject while merging, but first you always need to do this manually, as Gitlab never takes the title of the merge request as subject of the commit and secondly this is not exactly what we want to do.

To note, it would be very useful if one can just plug in new presets. Actually this doesn't seem to be possible?

stevemao:
The commit messages should still appear in the branch if you keep the merge commits. So the changelog should be generated from those commits (originally from another branch) but not merge commits. Also gitlab is currently not supported. #80 

It is possible to specify parserOpts, writerOpts etc. check the cli docs.

wzrdtales:
No they don't as we define the feature, fix etc. in the merge request not in the commits included in the merge requests.

stevemao:
So these commits are not technically merge commits but the message looks like a merge commits. 

The workflow or how you manage the commits is not the same as angular. I don't know if your workflow is common enough to be a preset. Again, you can always define all these configs in your own project.

stevemao:
So in your master branch all commits are like "merge branch bla..."?

I think there are so many ways to manage your src code especially with git you can go crazy with stuff like branches. The presets should just provide the commonly used ones.

wzrdtales:
@stevemao Yet I don't see any possibility to load a custom preset from the command line and it also doesn't seem like it is possible (unless I walk back many ../ times).
https://github.com/ajoslin/conventional-changelog/blob/master/index.js#L66

stevemao:
No. presets are a bunch of predefined options. You shouldn't/can't make a new preset but define all your options. The options are including `gitRawCommitsOpts`, `parserOpts` and `writerOpts` etc (this will be improved in the future #100).

wzrdtales:
I would really prefer to include my own preset, instead of writing a script and use gitRawCommitsOpts though. Including a custom preset would be also way cleaner in my mind.

Considering this pull request, there is at least some things to discuss about. For example the ember preset looks always for a merge request, similar to what we are doing.
https://github.com/ajoslin/conventional-changelog/blob/master/presets/ember.js#L11

If someone would want to use the ember conventions on gitlab, he wouldn't be able to. As merge request and pull request in gitlab are always titled in this scheme

    Merge branch 'feat/amazing' into 'develop'


And to come back to this

> I think there are so many ways to manage your src code especially with git you can go crazy with stuff like branches. The presets should just provide the commonly used ones.

I think at least the users should always be able to choose, whether they want to use only merge/pull requests to build the changelog, or search in the subject line only. And maybe it would be a solution to look out to put things like merge request subjects into generic functions, which can provide the different schemas from github, gitlab, bitbucket and so on.

stevemao:
> I would really prefer to include my own preset, instead of writing a script and use gitRawCommitsOpts though. Including a custom preset would be also way cleaner in my mind.

You would have to write the same amount of code if there was a way to define customised presets. The only difference is you could put gitRawCommitsOpts, parserOpts and writerOpts into one file. The difference is very trivial 

>  For example the ember preset looks always for a merge request, similar to what we are doing.
https://github.com/ajoslin/conventional-changelog/blob/master/presets/ember.js#L11

Because the ember project is using this way to manage the code. And the presets are made just for those popular projects. (And it's more likely the common use cases. I assume people who use the same conventional message would follow the same or similar workflow)

> I think at least the users should always be able to choose, whether they want to use only merge/pull requests to build the changelog, or search in the subject line only. 

How? 

Note that the merge messages should not related to git hosts. There might be another host that uses a completely different message. I would rather use the cli or a tool than just clicking the button on the website.

wzrdtales:
> You would have to write the same amount of code if there was a way to define customised presets. The only difference is you could put gitRawCommitsOpts, parserOpts and writerOpts into one file. The difference is very trivial 

This might seem trivial, but it is much more cleaner to ship a preset file in your repo and require it by the cli instead of having to define them all. But thats just my opinion.

wzrdtales:
> Note that the merge messages should not related to git hosts. There might be another host that uses a completely different message. I would rather use the cli or a tool than just clicking the button on the website.

This is correct, but bitbucket, gitlab and so on are very widespread, thus there are many users who are using the same schema. However I'm not against your opionion considering this, it is not possible to support everything. But yet I really think it is more friendly to the user to include custom presets in the same manner which the currently official presets are used via the command line.

stevemao:
It makes sense. If you can think of a good way it would be great :)

wzrdtales:
There are many ways.

One way would be to use the [resolve module](https://github.com/substack/node-resolve), if using this, one can look if in the local users project is a module installed, which is named like the package/preset the user is searching for and if it is not available, fallback to the project presets, I would require that a package must be named with a prefix, such as conventional-changelog-my-preset. That also allows users to share their custom presets.

Another way would be to use something like https://github.com/cowboy/node-findup-sync, to search for the root directory of the project starting from the current working directory. Which are indicated in this case by the .git folder or package.json. When this has been found the next step would be to search in the preset folder for the preset.

The next way would be to add an option for the user, `--custom` or something like this. This option would require a path which is used to search for the preset.

stevemao:
Oh, sorry. I didn't realise you were talking about customised presets. What's the difference between loading a bunch of opts files and one preset file? Also as I have already mentioned, the cli will be improved #100

stevemao:
I thought you meant having an option to use "merge" commits

wzrdtales:
@stevemao Having options for this as well :), but we just talked into to different directions thats right.

Ok let's first finish on the first question. The difference is:

* It is more user friendly
* You can ship your preset into multiple projects
* You can always refer to specific presets in your documentation
* Users can share their presets

Given my points, that it would be nice solution to make the user able to write his own publishable presets would incredibly increase the value of your module. It would increase the routines and development strategies supported, while spreading the support all over the community.
Also it is more straightforward to write a preset in the same way like they're done in the project. For my side, if I look at your documentation, I do not even notice that there would be any possibility to extend this module with the users settings and presets. Probably because there is no real documentation about the cli yet?
Also to note, we're a community of developers, thus many of us going straightforward through the source code and take a look how things are done if the documentation doesn't reveal the necessary details, and this leads directly to the presets.
And the big question is:
Why should the user use different CLI options to include a custom preset? There is nothing special about it, beside it is custom. And the first option that the user chooses is always to load another preset and not to use the parserOpt options and stuff.

wzrdtales:
@stevemao I will prepare an example PR the next days considering the merge commits.

stevemao:
Cool, just letting you know that it's probably not a simple task and I have no idea how to do it. So it might take some time to merge :p

stevemao:
This is configurable. Please try the master branch of the modules.

Issue #113, opened 2015-10-10
no docs on how to use `transform`


seiyria:
It's pretty frustrating to try to figure this out.

I've tried this:
```js
gulp.task('generate:changelog', function() {
  return changelog({
    releaseCount: 0,
    transform: function(commit, cb) {
      console.log(commit, cb);
      return cb(null, commit);
      // if(_.contains(commit.header, ['bump'])) return false;
    }
  })
    .pipe(fs.createWriteStream('CHANGELOG.md'));
});
```

I've also tried passing in `{ transform: ()... }` as the last param (`writerOpts`). I've tried returning true, returning false, returning callback(). All I'm trying to do is filter out some bad commits from my changelog.

[edit]
Some mild success doing it this way: 
```js
transform: function(commit, cb) {
      console.log(commit);
      if(_.contains(commit.header, ['bump'])) return cb(null);
      return cb(null, commit);
    }
```

and having it on the first options object. This merges everything into the latest tag though which is ... not desirable.

stevemao:
I think you need `writerOpts.transform`. PR welcome to improve the naming confusing :)

Idealy
> // it would be better to if `gitRawCommits` could spit out better formatted data
        // so we don't need to transform here

stevemao:
Or suggest better names.

seiyria:
Yeah, I'm not entirely sure. I couldn't even give a PR because I have no idea how to make it work as expected. That said, I just rebased my whole repo and started using the angular convention, so this isn't that big of a deal anymore.

Still, having some documentation on how to use `transform` would be a good idea.

stevemao:
First thing first, if someone could suggest better API names (there are three `transform`s) it would be great.

stalinkay:
Hi @stevemao Is it possible to use a custom clone of the angular preset with the CLI. I can get it to work by dropping it into the global conventional-changelog directory and simply specifying the name of my custom preset.

stevemao:
Use `--config`

stevemao:
Could you look into it and see what the problem is? If there's any bug please open a new issue. Thanks!

stalinkay:
@stevemao Ok. Cool.

PR #114, opened 2015-10-12
feat(preset): add dynamic preset loading

tamascsaba:
It possible load dynamic preset js.

Example:
```
conventional-changelog -p ./presets/custom -i CHANGELOG.md -w -r 0
conventional-changelog -p /foo/bar/presets/custom -i CHANGELOG.md -w -r 0
```
and the default behavior is work to (load conventional-changelog/presets/...):
```
conventional-changelog -p angular -i CHANGELOG.md -w -r 0
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ Returns a readable stream.
 
 ##### preset
 
-Type: `string` [Possible values](presets)
+Type: `string` [Default values](presets) or any valid absolute or relative path
 
 It's recommended to use a preset so you don't have to define everything yourself. The preset values can be overwritten.
 
--- a/index.js
+++ b/index.js
@@ -63,11 +63,11 @@ function conventionalChangelog(options, context, gitRawCommitsOpts, parserOpts, 
   var loadPreset = options.preset;
 
   if (loadPreset) {
-    var avaliblePresets = fs.readdirSync('./presets/');
     try {
+      var avaliblePresets = fs.readdirSync(__dirname + '/presets/');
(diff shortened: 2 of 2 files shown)


tamascsaba:
Upps, in node 0.10 path has no method parse :(

stevemao:
I think its more important to solve #113 and #100

tamascsaba:
Sorry, but I don't understand the couple whit this issues. It does not concern my use cases.

tamascsaba:
I will make comment this issues/discussions, as soon as possible.

stevemao:
Sorry if they are not obvious. In #100 I stated that the current CLI flags are not good and the purpose of preset is for people to load the preset of a popular project but not to customise it (if the flags are good enough we can config everything much easier).

tamascsaba:
Thanks for the answer. I understand it. You are right. My use case: Generate debian changelog, but templates are hard coded to presets. What do you think about: templateFolder flag, which whit h can you modify the default templates?

tamascsaba:
@stevemao ?? templateFolder flag??

stevemao:
a template flag might be good. As I said this really depends on #100. Would you like to work on #100 if you can?

stevemao:
The conversation here is a sub set of 
> Also we should probably remove --raw-commits-opts, --parser-opts and --writer-opts and expose useful options of them with each single flag. --context can still be a json file.

stevemao:
fixed by https://github.com/stevemao/conventional-changelog-core/commit/85fd9d9be52b88c6f79131e3095c0ef571a1b004

tamascsaba:
Nice, thx ☺

PR #115, opened 2015-10-15
docs(conventions/angular): add `docs` type prefix

Kaixhin:
Angular's [guidelines](https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#type) list `docs` as one of their compulsory set of type prefixes, and it is currently the only one listed there that is not documented here.

On a related note, perhaps it would be good to highlight any differences between conventional-changelog's interpretation and the actual Angular spec.


stevemao:
Thanks :)

PR #116, opened 2015-10-19
fix(options): fix losing parserOpts without preset

nmalaguti:
`preset.parserOpts` may be undefined if no preset was specified in the
options. `_.assign()` will return `undefined` if the first argument is
`undefined`.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -743,4 +743,25 @@ describe('conventionalChangelog', function() {
         done();
       }));
   });
+
+  it('should pass `parserOpts` to conventional-commits-parser', function(done) {
+    writeFileSync('test9', '');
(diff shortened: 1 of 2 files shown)


stevemao:
You need to provide tests to prove that this is actually fixing something :)

nmalaguti:
Test written :).

test/test.js
```diff
@@ -743,4 +743,25 @@ describe('conventionalChangelog', function() {
         done();
       }));
   });
+
+  it('should pass `parserOpts` to conventional-commits-parser', function(done) {
+    writeFileSync('test9', '');
+    shell.exec('git add --all && git commit -m"test9" -m"Release note: super release!"');
```
stevemao: remove ` -m"test9"`

stevemao:
Thanks. Merging.

PR #122, opened 2015-11-15
Own .editorconfig for templates and Angular templates fixes

estliberitas:
Hi guys! These two commits add `.editorconfig` to `templates/` folder and fix Angular preset templates so they generate less useless newline characters respectively. Hope you find this PR useful - it's not something new or cool but aims to make `conventional-changelog` output more accurate.


stevemao:
I prefer to merge .editorconfig with the root one.
Why do you think the newlines is a bug? What is it trying to fix?

stevemao:
Ref for .editorconfig https://github.com/stevemao/conventional-changelog-writer/blob/master/.editorconfig#L14-L16

estliberitas:
> Why do you think the newlines is a bug? What is it trying to fix?

stevemao:
I add the newlines on purpose to make the log easy to read.

estliberitas:
@stevemao Well, do you often read raw Markdown? As for me, yes, I've got no rendering in some project and then I saw generated file with things like `\n\n\n\n` - that's why I created PR. If this small addition is no good, let's close and forget )

stevemao:
I do. I found three `\n`s between versions make it easier to read. There are two `\n`s between sections and one `\n` between title and content.

estliberitas:
Okay then )

stevemao:
Do let me know if you find `\n` is wrong. rules should be pretty much my last comment :)

PR #131, opened 2015-12-31
lock git-semver-tags to 1.0.0

vladikoff:


stevemao:
git-semver-tags 1.1.1

PR #132, opened 2016-01-06
feat(changelog): Can specify a readable stream a commits source.

brettp:
We had a need to specify commits from a stream or file rather than using a git repo, so this PR adds an extension point for specifying the source of the commits.

CLI has been updated to add --from-file (-f) and --delimiter (-d) to specify a commits file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,6 +229,12 @@ $ npm install --global conventional-changelog
 $ conventional-changelog --help # for more details
 ```
 
+#### Generating logs to use with --from-file
+
+```sh
+$ git log --format="%B%n-hash-%n%H%n-gitTags-%n%d%n-committerDate-%n%ci%n------------------------ >8 ------------------------" > logs.txt
+$ conventional-changelog -f logs.txt
+```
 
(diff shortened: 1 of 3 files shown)


stevemao:
This is exactly why we broke this module into smaller pieces so you can assemble your own. The module only "Generate a changelog from git metadata" So it won't fit here. I was going to build another module which generates a changelog from another source.

brettp:
Do you have a start to the module you mention? There's currently a large amount of code duplication required if you only want to pass in a different source.

stevemao:
I've created https://github.com/stevemao/conventional-changelog-core.
Still a lot of work to be done with modulisation.
For now you have to duplicate the code unfortunately.

brettp:
Alright, thanks.

stevemao:
Thanks for this though. If you have an idea of how to modulise this better please let me know. Thanks.

stevemao:
@brettp Try https://github.com/stevemao/conventional-changelog-core/tree/master/lib

PR #146, opened 2016-04-08
add custom preset

Xiphe:
see https://github.com/Jimdo/conventional-changelog-custom

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ See the [conventional-changelog-core](https://github.com/stevemao/conventional-c
 
 ##### preset
 
-Type: `string` Possible values: `'angular', 'atom', 'codemirror', 'ember', 'eslint', 'express', 'jquery', 'jscs', 'jshint'`
+Type: `string` Possible values: `'angular', 'atom', 'codemirror', 'custom', 'ember', 'eslint', 'express', 'jquery', 'jscs', 'jshint'`
 
 It's recommended to use a preset so you don't have to define everything yourself. `options.config` will be set to this value.
 
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "conventional-changelog-atom": "^0.1.0",
     "conventional-changelog-codemirror": "^0.1.0",
     "conventional-changelog-core": "^1.3.0",
+    "conventional-changelog-custom": "^0.0.1",
     "conventional-changelog-ember": "^0.2.0",
     "conventional-changelog-eslint": "^0.1.0",
     "conventional-changelog-express": "^0.1.0",


stevemao:
Very interesting idea. Maybe we can put that in other section but not in the preset. Just a few things I notice:

1. The docs is not complete.
2. You made `changelogrc-spec`. However, all my `presets` are not following it and the `config` option is not expecting that spec. Maybe we can come up with a better name.

stevemao:
The spec is around angular convention (which is not really mentioned there). Maybe we could change the name to `angular-extended-changelogrc-spec` or something like that.

Xiphe:
Thanks for the feedback, really appreciate that :rocket: 

> 1. The docs is not complete.

What exactly are you missing? I could embed the docs from `changelogrc-spec` but that would lead to potential stale/diverged states

> 2. ...  However, all my presets are not following it and the config option is not expecting that spec ...

Thats actually part of the plan. I don't expect each preset to be `changelogrc`-compliant, that would somehow mitigate the "out of the box" solution they're providing.  
I think it's totally fine to use `conventional-changelog` settings that do not care about a `.changelogrc` file. I intentionally did not call it `.conventional-changelogrc` because it's meant to provide higher level configuration for all tools that work with these concepts (commitizen, semantic release).

> 2. ... the config option is not expecting that spec ...

As mentioned above, I don't think is necessary that `conventional-changelog-(core)` knows about the spec. I just want to have a config/preset that does that.

> 2. ... Maybe we can come up with a better name.

I agree. I thought about s.th. like `.commitstylerc`. Really open to suggestions here :)

> 3. The spec is around angular convention ...

Thats not entirely true. The spec lends wording from angular style but does not anticipate any specific format. I'm really open for renaming in order to clarify the concept but I don't want to tie the spec to any specific implementation.

The `conventional-changelog-custom` config on the other hand actually [extends `conventional-changelog-angular`](https://github.com/Jimdo/conventional-changelog-custom/blob/master/index.js#L4)

I thought about naming it `conventional-changelog-angular-custom`, but as for now I'd like to reduce/remove this dependency/connection in the future by allowing custom config for [conventional-commits-parser](https://github.com/conventional-changelog/conventional-commits-parser) in `.changelogrc` files.

Xiphe:
To give more context: We also just published a first version of https://github.com/Jimdo/cz-conventional-changelog-custom that also consumes `.changelogrc` files.

stevemao:
> The docs is not complete.

I think I was looking at https://www.npmjs.com/package/get-changelogrc-config

> Maybe we can come up with a better name.

I thought `conventional-changelog-angular-custom` was a really good name but

> but as for now I'd like to reduce/remove this dependency/connection in the future by allowing custom config for conventional-commits-parser in .changelogrc files.

Can you give me more details on that plan? Currently the config of `conventional-changelog` is very hard to write. Would it possible to follow the spec and produce presets like jQuery?

`.commitstylerc` sounds good but I wanna know more details on the spec.

Anyway, I really like your idea! Great work.

stevemao:
If I understand you correctly, `.changelogrc` can be consumed by `commitizen`, `semantic release` and `conventional-changelog`, but I guess it is possible to replace the current `config`. That means it may not be able to config presets such as `jQuery`. 

The name `custom` is definitely misleading I think.  `.commitstylerc` sounds better.

Xiphe:
Thanks for the feedback. I'll think about it and post and update soon

> Would it possible to follow the spec and produce presets like jQuery

That's the plan. I'll give that a shot asap

stevemao:
Hi @Xiphe, is there any updates?

Xiphe:
sorry, shifted my focus in the past months. We could close here or I can invite interested ppl to the projects I started.

Not sure if and when I'll find timt to continue the Idea

stevemao:
No problem. Let's keep the issue open. Feel free to invite people :)

bcoe:
@Xiphe we've migrate the main conventional-changelog repo to a mono-repo, so some updates will be necessary on this pull request -- did you ever find any folks interested in moving this work forward?

Xiphe:
Hey @bcoe thanks for the update. Afaik neither me nor others are interested in a continuation of this one.

Closing in favor of possible future approaches to this :)

stevemao:
I'm happy to leave this open :)

Issue #149, opened 2016-06-10
Release: Presets should be peer dependencies


DevVersion:
Currently, as presets are included by default. This is not desired in some projects.

Using `peerDepdencies` will allow developers to only include the most necessary presets.

stevemao:
Definitely something worth to be discussed.

stevemao:
I think I like the idea 😄

Tapppi:
Won't npm@2.x install all of them anyway and npm@3 print a ton of warnings for the non-installed peer dependencies though? I'd prefer not having a ton of warnings on each install, and propose a "plugin" approach:

1. This package doesn't depend on presets.
2. The presets have conventional-changelog as a peer dependency.
3. The preset is used by explicitly depending on it in the using package's `package.json` `dependencies` and giving the required package as the config option (or `preset`? but that seems strange). e.g. `conventionalChangelog({ config: require('conventional-changelog-angular' })`

The config options is overwritten with the preset anyway is 'preset' is used, so that doesn't change how it works, just how presets are included.

stevemao:
I think we should remove preset and only have `config`. if `config` is a string then load it as a preset.

Tapppi:
That seems logical, so the presets would be different packages and required by giving the package name instead of a `config` object to `conventionalChangelog`?

stevemao:
@Tapppi Probably both. But the idea is it should be only one `config`. we could override it.

hbetts:
I like the idea of making all presets peer dependencies.

Would someone be willing to take on this refactor for the community?

dangreen:
Presets are now installed explicitly and the core package has no preset dependencies; the original goal is met through this clearer model. Closing.

PR #156, opened 2016-10-30
docs: add conventional-changelog-angular-emoji

ellerbrock:
emoji support make you dance :)


stevemao:
Hi @ellerbrock thanks for the PR. However, this looks like a config of conventional-changelog rather than a related project. Maybe we can come up with a better solution after https://github.com/conventional-changelog/conventional-changelog/issues/149 is resolved 😄

bcoe:
closing since there hasn't been activity on this in quite some time; @ellerbrock feel free to re-open this discussion if this is something you'd like to still land in the README 👍

PR #167, opened 2017-03-06
feat: allow git-raw-commits to filter on specific paths

bcoe:
Allows the listing by `git-raw-commits` to be filtered to a specific path; this will allow a CHANGELOG to be generated for a specific package in `./packages/foo`.

I've created the pull against `lerna-semver-tags`, as this is all part of a project to add lerna support to conventional-changelog.


Tapppi:
@bcoe Should this point to master though? Or do you intend to land a 1, 2 punch with the fixes to lerna-semver-tags and this?

bcoe:
@Tapppi I accidentally forked from the wrong branch, but figured it wasn't the end of the world since we'd probably land the features close to each other.

PR #171, opened 2017-03-07
docs(badges): use svg for travis badge

JaKXz:


stevemao:
Thanks again :)

JaKXz:
np, easy wins :)

Issue #168, opened 2017-03-06
make conventional-changelog play nicely with lerna


bcoe:
## Goal

I would like to make conventional-changelog work as elegantly with lerna as the [standard-version](https://github.com/conventional-changelog/standard-version#standard-version) workflow on singular repos.

### In Independent Mode

* It would be nice if logic similar to [conventional-recommended-bump](https://github.com/conventional-changelog/conventional-recommended-bump) recommended version updates (rather than manual prompting).
* It would be nice to have a pretty CHANGELOG auto-generated for each module in `./package`.

### In Regular Mode

* It would be nice to have a pretty CHANGELOG auto-generated for each module in `./package`.

## Work To Do

- [x] scope `git log` to specific directory, so changes can be fetched for specific packages: https://github.com/conventional-changelog/conventional-changelog/pull/167
- [x] add support for lerna-format semver tags: https://github.com/conventional-changelog/conventional-changelog/pull/161
- [x] figure out an approach for hooking `conventional-changelog` into the lerna tagging process: https://github.com/lerna/lerna/issues/643
- [x] add support to conventional-changelog-core for lerna style commit tags (`pkg-name@version`).

CC: @hzoo, @Tapppi, @stevemao

Tapppi:
#167 and #161 should be a-ok to go, @bcoe has been doing awesome progress.

evocateur:
Oh geez, I would love to start re-using some of these packages (`git-semver-tags`, maybe more?) internally in `lerna`...

nicojs:
Cool! This couldn't come at a better time for us as we're migrating to lerna as we speak and already use the awesome conventional-changelog. 

When will this be released?

bcoe:
@nicojs would love the help testing; you can try things out by installing the canary build of lerna:

`npm i lerna@canary`.

nicojs:
@bcoe i will definitely do that later today. How does it work? We just run `grunt conventionalChangelog` using a `lerna run` job?

bcoe:
I'll document upstream, and perhaps write a bit of a post; currently it's
only enabled for independent mode (would love your thoughts on regular
mode, if this is what you're using) but! in independent mode you simply run:

lerna publish --conventional-commits

nicojs:
Cool! How can i test it without doing an actual npm release? Or just use a different npm-tag?

wizardzloy:
@nicojs 
```sh
lerna publish --skip-npm --conventional-commits
```
[...]

PR #173, opened 2017-03-09
feat: wire up lerna repository support

bcoe:
makes it possible to pass lerna configuration settings to `gitSemverTags`, and adds tests experimenting with the lerna repository support I've been working on.

see: #168 

## Neat!

```js
var conventionalChangelogCore = require('../conventional-changelog-core');

conventionalChangelogCore({
  lernaPackage: 'git-raw-commits',
  config: require('../conventional-changelog-angular')
}, {}, {
  path: './'
})
  .pipe(process.stdout); // or any writable stream
```

The above code seems to work like a charm.

```md
<a name="1.1.2"></a>
## [1.1.2](https://github.com/stevemao/git-semver-tags/compare/git-semver-tags@1.1.2...v1.1.2) (2017-03-09)


### Features

* add support for listing lerna style tags (project[@version](https://github.com/version)) (#161) ([b245f9d](https://github.com/stevemao/git-semver-tags/commit/b245f9d))
* migrate repo to lerna mono-repo ([793e823](https://github.com/stevemao/git-semver-tags/commit/793e823))
```

## Todo

- [x] if no commits are found matching `lernaPackage` a CHANGELOG is generated with all git commits 🐞

### DIFF
--- a/packages/conventional-changelog-core/lib/merge-config.js
+++ b/packages/conventional-changelog-core/lib/merge-config.js
@@ -20,10 +20,21 @@ var _ = require('lodash');
 var rhosts = /github|bitbucket|gitlab/i;
 var rtag = /tag:\s*[v=]?(.+?)[,\)]/gi;
 
+function semverTagsPromise(options) {
+  var semverTagsDeferred = Q.defer();
+  gitSemverTags(function(err, result) {
+    if (err) {
+      semverTagsDeferred.reject(err);
+    } else {
+      semverTagsDeferred.resolve(result);
+    }
+  }, {lernaTags: !!options.lernaPackage, package: options.lernaPackage});
+  return semverTagsDeferred.promise;
+}
+
 function mergeConfig(options, context, gitRawCommitsOpts, parserOpts, writerOpts) {
   var configPromise;
   var pkgPromise;
-  var semverTagsPromise;
   var gitRemoteOriginUrlPromise;
 
   context = context || {};
@@ -53,7 +64,8 @@ function mergeConfig(options, context, gitRawCommitsOpts, parserOpts, writerOpts
       }
 
       cb(null, commit);
-    }
+    },
+    lernaPackage: null
   }, options);
 
   options.warn = options.warn || options.debug;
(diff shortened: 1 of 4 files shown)


packages/conventional-changelog-core/lib/merge-config.js
```diff
@@ -20,10 +20,21 @@ var _ = require('lodash');
 var rhosts = /github|bitbucket|gitlab/i;
 var rtag = /tag:\s*[v=]?(.+?)[,\)]/gi;
 
+function semverTagsPromise(options) {
+  var semverTagsDeferred = Q.defer();
```
stevemao: Can we use the `Promise` api instead of `defer()`? [Reason](https://github.com/petkaantonov/bluebird/wiki/Promise-anti-patterns#the-deferred-anti-pattern)

stevemao: Also packages like `gitSemverTags` was written long time ago. I think now they could be changed to return a promise.

stevemao:
Great work :) The bug should be easy to fix since you've already added the `path` option in `gitRawCommits` I guess :)

packages/conventional-changelog-core/lib/merge-config.js
bcoe: @stevemao @Tapppi were you thinking:

```js
new Promise(resolve, reject) {}
```

This is what I usually use, but wasn't sure what's supported by `q`.

Tapppi: Has `conventional-changelog-core` already dropped node `<v4` support? Then it would be nice to promisify things.. Or do it anyway with a major bump :smile: Anyway, I agree, defer is not very nice here.

stevemao: It's
```js
Q.Promise(function(resolve, reject, notify) { })
```

And yes to @Tapppi. But it's also out of scope here :)

bcoe:
@stevemao @Tapppi dug a bit further, I believe the bug is actually things behaving as expected; if you haven't tagged your first lerna-format tags yet, then one would expect the whole git history to be included -- I was finding things to be behaving strangely because I wasn't providing the `path` variable.

PR #174, opened 2017-03-09
feat(options): add option for scoped packages

yoitsro:
This allows for scoped preset packages to be used. Before, only non-scoped packages could be used as a preset, which wasn't great for packages scoped under an organisation's scope name.

### DIFF
--- a/packages/conventional-changelog/index.js
+++ b/packages/conventional-changelog/index.js
@@ -4,11 +4,19 @@ var conventionalChangelogCore = require('conventional-changelog-core');
 function conventionalChangelog(options, context, gitRawCommitsOpts, parserOpts, writerOpts) {
   options.warn = options.warn || function() {};
 
+  var scope = '';
+
+  if (options.scope) {
+    scope = options.scope.toLowerCase() + '/';
+  }
+
   if (options.preset) {
     try {
-      options.config = require('conventional-changelog-' + options.preset.toLowerCase());
+      options.config = require(scope + 'conventional-changelog-' + options.preset.toLowerCase());
     } catch (err) {
-      options.warn('Preset: "' + options.preset + '" does not exist');
+      var errMessage = 'Preset: "' + options.preset + '" does not exist';
+      errMessage += scope !== '' ?  ' under scope "' + options.scope + '"' : '';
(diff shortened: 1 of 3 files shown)


stevemao:
Thanks @yoitsro for the PR. But I don't think this fits the project. Can you join https://github.com/conventional-changelog/conventional-changelog/pull/146 to discuss?

Tapppi:
@yoitsro is there a reason you cant use:
```
const config = require('@org/conventional-changelog-custom-preset');
conventionalChangelog({config})...
```

@stevemao I was hoping to allow setting preset in `standard-version` in https://github.com/conventional-changelog/standard-version/issues/135#issuecomment-269008679 as a way to allow users to fix their edge cases without making `standard-version` bloated. If this is not in the scope for `conventional-changelog`, should we load the config in `standard-version`)

stevemao:
> @yoitsro is there a reason you cant use:
> 
> const config = require('@org/conventional-changelog-custom-preset');
> conventionalChangelog({config})...

Exactly what I mean. You can config `conventionalChangelog` now without using preset. Presets should be similar to the `jscs` presets, which shouldn't be too interesting but just some basic defaults.

I think #146 provides a very clean solution to not bloat `standard-version`.

stevemao:
We can use `standard-changelog` to handle this. It should allow angular convention like commit messages and transforms all the config to feed `conventional-changelog`. Then `standard-version` can provide simple config to `standard-changelog`.

yoitsro:
@Tapppi @stevemao yeah, I've just done that and it's great. It's really not that obvious though.

As a new user, you land [at the readme](https://github.com/conventional-changelog/conventional-changelog) which only halfway through mentions [the actual library](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog) which eventually leads you to documentation about the [options available](https://github.com/conventional-changelog-archived-repos/conventional-changelog-core). But as a new user, it's not entirely obvious that the presets encompass these options 🙈.

Anyway, maybe this will make it more obvious: https://github.com/conventional-changelog/conventional-changelog/pull/177

I'll close this PR for now :)

Thank you for your time and effort with this! It is appreciated.

PR #175, opened 2017-03-10
feat: update CLI tools to support lerna tags

bcoe:
for easier testing, add support for lerna-style CHANGELOG generation to `conventional-changelog-cli` and `standard-changelog`.

### DIFF
--- a/packages/conventional-changelog-cli/cli.js
+++ b/packages/conventional-changelog-cli/cli.js
@@ -45,7 +45,9 @@ var cli = meow({
     '  -n, --config              A filepath of your config script',
     '                            Example of a config script: https://github.com/conventional-changelog/conventional-changelog-angular/blob/master/index.js',
     '',
-    '  -c, --context             A filepath of a json that is used to define template variables'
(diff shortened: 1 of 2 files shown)


bcoe:
note that coverage dropped because coveralls has been rolling out branch coverage.

stevemao:
Previously I only exposed options of `conventionalChangelog` as flags and options for other submodules are in the `config` object. `commitPath` is an option of `gitRawCommitsOpts` so just note that this is a bit inconsistent once the PR is merged. But I think maybe we should expose all the important options as flags for conveniency. LGTM :)

bcoe:
@stevemao I think this might be worthwhile; I was picturing for lerna -- in line with how it's implemented other features [like git](https://github.com/lerna/lerna/blob/master/src/GitUtilities.js) -- that we could potentially just shell out to our conventional-changelog CLI tools.

It's easy to add tests for this in the lerna codebase, and cuts down on the # of top-level dependencies that we need to add to lerna itself.

PR #176, opened 2017-03-10
feat: pulls in conventional-recommended-bump adding support for lerna

bcoe:
in this pull request I import `conventional-recommended-bump` into our monorepo, and add support for recommending a bump based on lerna style tags.


stevemao:
dotfiles like `.gitattributes` could be moved to the trunk? But it doesn't prevent this PR getting merged. Great work :)

bcoe:
@stevemao good call, I will fix this.

Tapppi:
This is awesome @bcoe ! ❤️

PR #179, opened 2017-03-12
feat(context): default currentTag may not prefix with v

stevemao:
Before this PR, it always add a leading "v" to version number.

After this PR, based on the previous commit tag, it may add or remove the leading "v".

Closes https://github.com/conventional-changelog-archived-repos/conventional-changelog-core/pull/15
Fixes https://github.com/conventional-changelog-archived-repos/conventional-changelog-core/issues/10

### DIFF
--- a/packages/conventional-changelog-core/lib/merge-config.js
+++ b/packages/conventional-changelog-core/lib/merge-config.js
@@ -155,6 +155,8 @@ function mergeConfig(options, context, gitRawCommitsOpts, parserOpts, writerOpts
         context.packageData = pkg;
       }
 
+      context.version = context.version || '';
+
       if (tagsObj.state === 'fulfilled') {
         gitSemverTags = context.gitSemverTags = tagsObj.value;
         fromTag = gitSemverTags[options.releaseCount - 1];
@@ -230,6 +232,7 @@ function mergeConfig(options, context, gitRawCommitsOpts, parserOpts, writerOpts
 
       writerOpts = _.assign({
           finalizeContext: function(context, writerOpts, filteredCommits, keyCommit, originalCommits) {
+            var currentTag;
             var firstCommit = originalCommits[0];
             var lastCommit = originalCommits[originalCommits.length - 1];
             var firstCommitHash = firstCommit ? firstCommit.hash : null;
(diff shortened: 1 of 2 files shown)


stevemao:
@SteveVanOpstal this should fix your problem.

packages/conventional-changelog-core/lib/merge-config.js
```diff
@@ -264,9 +267,27 @@ function mergeConfig(options, context, gitRawCommitsOpts, parserOpts, writerOpts
                 } else {
                   context.currentTag = context.currentTag || firstCommitHash;
                 }
-              } else {
-                if (!context.currentTag) {
-                  context.currentTag = options.lernaPackage ? options.lernaPackage + '@' + context.version : 'v' + context.version;
+              } else if (!context.currentTag) {
```
bcoe: not a blocker; but there's gettin to be enough logic here, it might be worth pulling this into a helper ... guessNextTag()?

Tapppi:
Appveyor got some npm install error, wouldn't mind it, since the duplicate went through.

HACK21:
Is there any plan to add configuration `tag prefix`  into `context` ?

We have a lot of projects, they are all tagged with `publish/1.0.2` etc

stevemao:
@HACK21 This is already configurable via the writer.

HACK21:
@stevemao I'm using [conventional-changelog-cli](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-cli) generate changelog. 

When I create tag with prepfix  `publish/` , it always include all history commits. When I tag with prefix `v`, it goes right. I try to find the configuration like `prefixTag` in [conventional-changelog-core](https://github.com/conventional-changelog-archived-repos/conventional-changelog-core#conventionalchangelogcoreoptions-context-gitrawcommitsopts-parseropts-writeropts) but I failed.

stevemao:
You need to config [generateOn](https://github.com/conventional-changelog-archived-repos/conventional-changelog-writer#generateon), this is advanced config so you need to use a script via [--config](https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-cli/cli.js#L45)

PR #187, opened 2017-05-11
[WIP] feat(link-compare): generate compare url and simplify templates

### LINKED ISSUE #185: Bitbucket Compare Seperator
Wondering if its possible to add a "compare" param to the hosts files and expose that to the templates as github uses `...` for their compare links but bitbucket uses `%0D` which is a carriage return character.

Because of that the compare links don't function on Bit Bucket.

```
# Doesnt work on bit bucket
https://bitbucket.org/<user>/<repo>/compare/@my/npm-module@0.0.6...@my/npm-module@1.0.0
# Works on Bit Bucket
https://bitbucket.org/<user>/<repo>/compare/@my/npm-module@0.0.6%0D@my/npm-module@1.0.0
```

So i'm suggesting adding something like `{{@root.hostCompare}}`

```hbs
<a name="{{version}}"></a>
{{#if isPatch}}##{{else}}#{{/if}} {{#if @root.linkCompare}}[{{version}}]({{@root.host}}/{{#if @root.owner}}{{@root.owner}}/{{/if}}{{@root.repository}}/compare/{{previousTag}}
{{@root.hostCompare}} // added line breaks to make it clear where this is
{{currentTag}}){{else}}{{version}}{{/if}}{{#if title}} "{{title}}"{{/if}}{{#if date}} ({{date}}){{/if}}
```

admosity:
- generate compare url when there are tags
- added test for bitbucket host
- updated templates to use `linkCompare` compare directly

Improves #80 
Fixes #185, conventional-changelog/standard-version#142

Didn't go the template editing route because of:
https://github.com/wycats/handlebars.js/issues/927
and
http://chrismontrois.net/2016/01/30/handlebars-switch/

Prior work:
https://github.com/conventional-changelog-archived-repos/conventional-changelog-core/pull/21

### DIFF
--- a/packages/conventional-changelog-core/lib/merge-config.js
+++ b/packages/conventional-changelog-core/lib/merge-config.js
@@ -32,6 +32,39 @@ function semverTagsPromise(options) {
   });
 }
 
+/**
+ * Generates a compare url on the respective repository host
+ *
+ * @param {any} context the same context object as the one that's used in the templates
+ * @returns {string} compare url for gitlab | github | bitbucket
+ */
+function generateCompareUrl(context) {
+  var generatedUrl = '';
+  if (context.repository) {
+    if (context.host) {
+      generatedUrl += context.host + '/';
+    }
+    if (context.owner) {
+      generatedUrl += context.owner + '/';
+    }
+    generatedUrl += context.repository;
+  } else {
+    generatedUrl = context.repoUrl;
+  }
+  switch (context.type) {
+    case 'gitlab':
+    case 'github':
+      generatedUrl += '/compare/' + context.previousTag + '...' + context.currentTag;
+      break;
+    case 'bitbucket':
+      generatedUrl += '/compare/' + context.currentTag + '..' + context.previousTag;
+      break;
+    default:
+      generatedUrl += '/compare';
+  }
+  return generatedUrl;
+}
+
 function guessNextTag(previousTag, version) {
   if (previousTag) {
     if (previousTag[0] === 'v' && version[0] !== 'v') {
@@ -212,7 +245,8 @@ function mergeConfig(options, context, gitRawCommitsOpts, parserOpts, writerOpts
 
           context = _.assign({
             issue: hostOpts.issue,
-            commit: hostOpts.commit
+            commit: hostOpts.commit,
+            type: type // add extracted type to context
           }, context);
         } else {
           options.warn('Host: "' + context.host + '" does not exist');
@@ -296,7 +330,7 @@ function mergeConfig(options, context, gitRawCommitsOpts, parserOpts, writerOpts
             }
 
             if (!_.isBoolean(context.linkCompare) && context.previousTag && context.currentTag) {
-              context.linkCompare = true;
+              context.linkCompare = generateCompareUrl(context);
             }
 
             return context;
--- a/packages/conventional-changelog-core/test/test.js
+++ b/packages/conventional-changelog-core/test/test.js
@@ -492,6 +492,27 @@ describe('conventionalChangelogCore', function() {
     }));
   });
 
+  it('should generate correct compare url for bitbucket', function(done) {
+    preparing(18);
+
+    conventionalChangelogCore({
+      releaseCount: 0,
+      pkg: {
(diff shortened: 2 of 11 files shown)


packages/conventional-changelog-core/lib/merge-config.js
```diff
@@ -32,6 +32,39 @@ function semverTagsPromise(options) {
...
 
+/**
+ * Generates a compare url on the respective repository host
+ *
+ * @param {any} context the same context object as the one that's used in the templates
+ * @returns {string} compare url for gitlab | github | bitbucket
+ */
+function generateCompareUrl(context) {
+  var generatedUrl = '';
+  if (context.repository) {
```
admosity: Logic extracted from hbs templates

packages/conventional-changelog-core/test/test.js
```diff
@@ -492,6 +492,27 @@ describe('conventionalChangelogCore', function() {
...
 
+  it('should generate correct compare url for bitbucket', function(done) {
+    preparing(18);
+
+    conventionalChangelogCore({
+      releaseCount: 0,
+      pkg: {
+        path: __dirname + '/fixtures/_bitbucket-host.json'
+      },
+      config: require('../../conventional-changelog-angular')
```
admosity: Referenced the local changelog-angular

stevemao:
Thanks for the PR. Could you have a look at https://github.com/conventional-changelog-archived-repos/conventional-changelog-core/pull/21#issuecomment-263088224?

admosity:
I ran through that thread as well. I'm in the opinion of not adding helpers since most of the logic for generating the `context` is within `merge-config.js` and it seems that the compare url was the outlier here compared to the rest of the fields and their usage in the hbs templates.

I'm also sort of new to the library, so I'm not too sure of all the capabilities that conventional changelog provides. It looks like you can feed in your own handlebars template and in that case it might be reasonable to provide helpers for the end user to use.

admosity:
@stevemao ? Anything I should adjust here or think about here? I tried posting this in the other thread as well.

stevemao:
Sorry @admosity I haven't got time for this. I feel like the current API of writer is not good. Adding feature like this should be easy but it's not. See https://github.com/conventional-changelog-archived-repos/conventional-changelog-writer/issues/20

packages/conventional-changelog-core/lib/merge-config.js
```diff
@@ -32,6 +32,39 @@ function semverTagsPromise(options) {
...
 }
 
+/**
+ * Generates a compare url on the respective repository host
+ *
+ * @param {any} context the same context object as the one that's used in the templates
+ * @returns {string} compare url for gitlab | github | bitbucket
+ */
+function generateCompareUrl(context) {
+  var generatedUrl = '';
```
hbetts: Could you switch to `let` over `var`, and switch to template literals over single quotes?

packages/conventional-changelog-core/lib/merge-config.js
```diff
@@ -212,7 +245,8 @@ function mergeConfig(options, context, gitRawCommitsOpts, parserOpts, writerOpts
 
           context = _.assign({
             issue: hostOpts.issue,
-            commit: hostOpts.commit
```
hbetts: I wish the style rules for this repository allowed for trailing commas so that these lines wouldn't need to be touched to add that comma for a new line underneath. (Just my thoughts on the matter. Nothing actually needs to change here)

packages/conventional-changelog-core/test/test.js
hbetts: Is this the only location in `conventional-changelog-core` that we test for a valid `compare` link?

Should this be extended to also test for GItLab, GitHub, and a unknown SCM platform?

admosity:
Didn't know there was movement here. @hbetts I'll get around to reviewing what you've brought up soon.

hbetts:
Cross referencing #185

hbetts:
I am closing this pull request for now. Please feel free to re-submit the pull request if you have time to make progress on this work.

kceb:
@admosity did you get a chance to look into this?

Issue #165, opened 2017-03-06
add a files stanza to each package.json, to reduce package size


bcoe:
we should make sure there's an appropriate `files` stanza added to each `package.json` to reduce repo size.

see: https://github.com/conventional-changelog-archived-repos/conventional-changelog-core/pull/22

thejameskyle:
Done in https://github.com/conventional-changelog/conventional-changelog/pull/211

bcoe:
@thejameskyle thanks 👍 really appreciate your help on this project.

PR #198, opened 2017-06-08
chore(writer): add .npmignore to ignore test folder

stevemao:
I feel like adding a file array would be harder to maintain for this package. We have a few folders and files including `lib/`, `templates/`, `index.js` and `cli.js`. One might forget to update it.

Ref #165


Andarist:
@bcoe @stevemao any update on this?

Andarist:
@stevemao thanks for merging in! Any ETA when this hits npm publication?

PR #199, opened 2017-06-09
docs(README): add worflow with npm version

micky2be:
Not only that some people like myself might rather use `npm version` to bump their version.
I also believe it makes the workflow even simpler this way.

Closes conventional-changelog-archived-repos/conventional-changelog-cli#11

### DIFF
--- a/packages/conventional-changelog-cli/README.md
+++ b/packages/conventional-changelog-cli/README.md
@@ -42,7 +42,6 @@ To fully customize the tool, please checkout [conventional-changelog](https://gi
 - https://github.com/karma-runner/karma/blob/master/CHANGELOG.md
 - https://github.com/btford/grunt-conventional-changelog/blob/master/CHANGELOG.md
 
-
 ## Recommended workflow
 
 1. Make changes
@@ -57,10 +56,33 @@ To fully customize the tool, please checkout [conventional-changelog](https://gi
 
 The reason why you should commit and tag after `conventionalChangelog` is that the CHANGELOG should be included in the new release, hence `gitRawCommitsOpts.from` defaults to the latest semver tag.
 
-If you use `npm version`, it auto tags immediately after changing the version in package.json. In such case, you might want to specify the version manually and generate the changelog before `npm version`.
 
-Please use this [gist](https://gist.github.com/stevemao/280ef22ee861323993a0) to make a release or change it to your needs.
+## Recommended workflow using `npm version`
+Using the npm scripts to our advantage.
+With the following hooks:
+```json
+{
+  "scripts": {
+    "version": "conventional-changelog -p angular -i CHANGELOG.md -s -r 0 && git add CHANGELOG.md"
+  }
+}
+```
+You could follow the following workflow 
 
+1. Make changes
+2. Commit those changes
+3. Pull all the tags
+4. Run the [`npm version [path|minor|major]`](https://docs.npmjs.com/cli/version) command
+5. Push
+
+You could optionaly add a `preversion` script to package your project or running a full suit of test.
+And a `postversion` script to clean your system and push your release and tags.
+
+By adding a `.npmrc` you could also automate your commit message and set your tag prefix as such:
+```bash
+tag-version-prefix=""
+message="chore(release): %s :tada:"
+```
 
 ## Why
 


micky2be:
Do I need to do any modification here?

packages/conventional-changelog-cli/README.md
```diff
@@ -57,10 +56,33 @@ To fully customize the tool, please checkout [conventional-changelog](https://gi
...
+  "scripts": {
+    "version": "conventional-changelog -p angular -i CHANGELOG.md -s -r 0 && git add CHANGELOG.md"
+  }
+}
+```
+You could follow the following workflow 
 
+1. Make changes
+2. Commit those changes
+3. Pull all the tags
```
hbetts: Why aren't all the tags already available?

packages/conventional-changelog-cli/README.md
```diff
@@ -57,10 +56,33 @@ To fully customize the tool, please checkout [conventional-changelog](https://gi
...
-If you use `npm version`, it auto tags immediately after changing the version in package.json. In such case, you might want to specify the version manually and generate the changelog before `npm version`.
 
-Please use this [gist](https://gist.github.com/stevemao/280ef22ee861323993a0) to make a release or change it to your needs.
+## Recommended workflow using `npm version`
+Using the npm scripts to our advantage.
+With the following hooks:
+```json
+{
+  "scripts": {
+    "version": "conventional-changelog -p angular -i CHANGELOG.md -s -r 0 && git add CHANGELOG.md"
```
hbetts: Could this be `postversion` for clarity? (Making it obvious that it occurs _after_ the behavior of `version`)

packages/conventional-changelog-cli/README.md
```diff
@@ -57,10 +56,33 @@ To fully customize the tool, please checkout [conventional-changelog](https://gi
...
+  }
+}
+```
+You could follow the following workflow 
 
+1. Make changes
+2. Commit those changes
+3. Pull all the tags
+4. Run the [`npm version [path|minor|major]`](https://docs.npmjs.com/cli/version) command
+5. Push
```
hbetts: I think you still have to execute `git commit` to commit the modified `CHANGELOG.md` file.

packages/conventional-changelog-cli/README.md
```diff
@@ -57,10 +56,33 @@ To fully customize the tool, please checkout [conventional-changelog](https://gi
...
+
+You could optionaly add a `preversion` script to package your project or running a full suit of test.
+And a `postversion` script to clean your system and push your release and tags.
+
+By adding a `.npmrc` you could also automate your commit message and set your tag prefix as such:
```
hbetts: Where do you add the `.npmrc` file?

packages/conventional-changelog-cli/README.md
micky2be: `nvm version` run `git commit` for you, that the whole goal here

packages/conventional-changelog-cli/README.md
micky2be: I can add a link to the npmrc documentation here
(https://docs.npmjs.com/files/npmrc)

packages/conventional-changelog-cli/README.md
micky2be: locally, you usually don't have all the tags.
unless you work alone and tag everything by yourself

packages/conventional-changelog-cli/README.md
micky2be: It will defy the purpose and will need to create an other commit.
The goal here is to update the version, generate the changelog, then commit all

micky2be:
As for `yarn` using `yarn version` instead of `npm version` should do the same as it's just an alias I believe

packages/conventional-changelog-cli/README.md
hbetts: The `npm version` documentation seems to imply that the author of a `version` script should add any files using `git commit` - `Again, scripts should explicitly add generated files to the commit using git add.`

So `npm version` may be adding files, but it seems like npm expects users to do that themselves.

micky2be: yes need to be explicitly added (not committed), that's what I'm doing in the `version` hook.

That's what I use in a daily bases on several repos.

packages/conventional-changelog-cli/README.md
hbetts: Ah, so taking advantage of the fact that `npm` does not run `git commit` until the `version` script has completed.

Thank you for the clarification.

`postversion` still seems clearer, but taking advantage of `npm`'s behavior for the `version` script is a clear win for usability.

packages/conventional-changelog-cli/README.md
hbetts: Thank you for the clarification @micky2be!

hbetts:
:+1: Looks good to me!

stevemao:
Merged thanks :)

PR #200, opened 2017-06-13
docs: fix typo in `lernaPackage` option docs

no23reason:
There is a typo the `lernaPackage` option docs. It states that the structure should be `./package/foo-package` but according to lerna docs and also the [`conventional-changelog-core` unit tests](https://github.com/conventional-changelog/conventional-changelog/blob/431598a3b12a404aee63f8ab3e51263121cde485/packages/conventional-changelog-core/test/test.js#L1222) it should be `./packages/foo-package`


stevemao:
Thanks :)

PR #206, opened 2017-07-11
fix: find package.json from cwd upwards

marionebl:
* smarten up logic to find a "root" package json a bit
* use pkg-up to find "root"
* use load-json-file for nice error messages
* move module scope procedure to getPkg
* handle missing package.json without logged warning
* propagate possible json errors

### DIFF
--- a/packages/conventional-changelog-angular/index.js
+++ b/packages/conventional-changelog-angular/index.js
@@ -1,19 +1,10 @@
 'use strict';
 var compareFunc = require('compare-func');
+var gufg = require('github-url-from-git');
+var readPkgUp = require('read-pkg-up');
 var Q = require('q');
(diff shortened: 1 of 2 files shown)


stevemao:
Thanks for the PR! Could you use [`read-pkg-up`](https://www.npmjs.com/package/read-pkg-up)?

stevemao:
Also the commit message should probably be:
`feat(angular): find package.json from cwd upwards`

marionebl:
Sure, adapted the changes according to your suggestions.

marionebl:
Could you retrigger https://ci.appveyor.com/project/stevemao/conventional-changelog/build/1.0.126? npm seems to have acted up on this build.

PR #216, opened 2017-08-08
feat(angular): use the context for getting the package.json and the url

dlmr:
This will make result in more flexibility since we now use the logic in `conventional-changelog-core` which will fallback to use `read-pkg-up`.

https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-core#pkg

Related PR #206

### DIFF
--- a/packages/conventional-changelog-angular/index.js
+++ b/packages/conventional-changelog-angular/index.js
@@ -1,7 +1,6 @@
 'use strict';
 var compareFunc = require('compare-func');
 var gufg = require('github-url-from-git');
-var readPkgUp = require('read-pkg-up');
 var Q = require('q');
 var readFile = Q.denodeify(require('fs').readFile);
 var resolve = require('path').resolve;
@@ -18,8 +17,8 @@ var parserOpts = {
   revertCorrespondence: ['header', 'hash']
 };
 
-function issueUrl() {
-  var pkg = readPkgUp.sync().pkg;
+function issueUrl(context) {
+  var pkg = context.packageData;
 
   if (pkg && pkg.repository && pkg.repository.url && ~pkg.repository.url.indexOf('github.com')) {
     var gitUrl = gufg(pkg.repository.url);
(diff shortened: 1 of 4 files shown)


stevemao:
@marionebl is there anything missing here?
@dlmr I think this also makes it easy to write a unit test. Could you add a test for it? Thanks.

dlmr:
Added two tests, one verifying that it works without any custom configuration and one where a path is specified.

Writing this tests highlighted a smaller inconstancy, it looks like most urls in the  `package.json` for the projects are incorrect and are pointed to where the projects where located before converting them into the monorepo that we have today. 

For example is the repository url for `conventional-changelog-angular` is `https://github.com/conventional-changelog/conventional-changelog-angular` when it should be `https://github.com/conventional-changelog/conventional-changelog`.

dlmr:
I looked a little more into this and I'm thinking we probably should do this in another way.

The reason for this is that we right now seem to do double work since we already use `get-pkg-repo` in `conventional-changelog-core` and set the result on the `context` here: https://github.com/conventional-changelog/conventional-changelog/blob/431598a3b12a404aee63f8ab3e51263121cde485/packages/conventional-changelog-core/lib/merge-config.js#L166-L173

@stevemao If you think it sounds like a good idea I will change the code to use what we have on the context already, primarily `context.repoUrl`. I also think it is a good idea to use `context.host` here: https://github.com/conventional-changelog/conventional-changelog/blob/431598a3b12a404aee63f8ab3e51263121cde485/packages/conventional-changelog-angular/index.js#L90

This would make it a bit more generic, supporting some instances of GitHub Enterprise.

stevemao:
@dlmr I'm not following but please go ahead and do whatever you think is right 😄

dlmr:
@stevemao I have now implemented what I was talking about in #217 that will replace this PR if accepted 🙂

dlmr:
@stevemao Is there a schedule when this will be released? 🙂

PR #217, opened 2017-08-10
feat(angular): use the context for getting the repository and host urls

dlmr:
This will result in more flexibility since we now use the logic already existing in `conventional-changelog-core` and at the same time make the code simpler. The logic now matches the one already present in [the templates for inserting a link to the commit](https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-angular/templates/commit.hbs#L9-L21).

This PR supersedes and replaces #216

### DIFF
--- a/packages/conventional-changelog-angular/index.js
+++ b/packages/conventional-changelog-angular/index.js
@@ -1,7 +1,5 @@
 'use strict';
 var compareFunc = require('compare-func');
-var gufg = require('github-url-from-git');
-var readPkgUp = require('read-pkg-up');
 var Q = require('q');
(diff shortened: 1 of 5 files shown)


dlmr:
When do you think that we will cut a new release with this @stevemao? No real rush at all just wondering if I should use a local fork in a project of mine in the meantime or hold out a little for this to be available.

Thanks for the quick reviews and feedback!

stevemao:
@dlmr could you have a look at https://github.com/conventional-changelog/conventional-changelog/issues/218?

PR #219, opened 2017-08-14
fix(angular): smarter username detection

### LINKED ISSUE #218: Angular preset: disable or avoid faulty username replace
Hello there 👋 

With the [current implement](https://github.com/conventional-changelog/conventional-changelog/blob/e0d7922ea05bd67e3183ea4a75ed38b95e3b7948/packages/conventional-changelog-angular/index.js#L83), a fix comment like `semantic-release@7.0.0 ` gets parsed to `semantic-release[@7](https://github.com/7).0.0`, see an example at https://github.com/gr2m/semantic-release-test/releases/tag/v2.0.2

This will happen pretty often once I release the new semantic-release version, as many people use Greenkeeper & semantic-release together, and each time a non-dev dependency gets updated, the problem will occur.

I wonder if there is a way to prevent that from happening, maybe by checking if the username exists? We could at least make sure that the first letter after the `@` is not a number, but there will still be problems with scoped packages, and I don’t know how to identify these, because something like `@semantic-release/generate-release-notes` could as well be a team on GitHub ...

Any ideas?

dlmr:
Solves issues where incorrect links would be inserted when describing the version of something, like `npm@5`. Now a space is required before a username to be replaced with a link to the profile. 

Closes #218 

---

_This PR has been based on the changes done in #217 and will be rebased on master after it has been merged._

### DIFF
--- a/packages/conventional-changelog-angular/index.js
+++ b/packages/conventional-changelog-angular/index.js
@@ -70,7 +70,7 @@ var writerOpts = {
       }
       if (context.host) {
         // User URLs.
-        commit.subject = commit.subject.replace(/@([a-zA-Z0-9_]+)/g, '[@$1](' + context.host + '/$1)');
+        commit.subject = commit.subject.replace(/\B@([a-z0-9](?:-?[a-z0-9]){0,38})/g, '[@$1](' + context.host + '/$1)');
       }
     }
 
--- a/packages/conventional-changelog-angular/test/test.js
+++ b/packages/conventional-changelog-angular/test/test.js
@@ -50,6 +50,9 @@ betterThanBefore.setups([
   },
   function() {
     gitDummyCommit(['feat(*): implementing #5 by @dlmr', ' closes #10']);
+  },
+  function() {
+    gitDummyCommit(['fix: use npm@5 (@username)']);
   }
 ]);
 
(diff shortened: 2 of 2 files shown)


stevemao:
@dlmr I have merged https://github.com/conventional-changelog/conventional-changelog/pull/217. Thanks!

dlmr:
Rebased on the PR on master.

dlmr:
Changed the regexp used to match username to something a little smarter, [as mentioned in this comment](https://github.com/conventional-changelog/conventional-changelog/issues/218#issuecomment-322135472).

stevemao:
Thanks @dlmr 
@gr2m could you try this?

stevemao:
> I wonder if we could maybe make it for usernames only, and simply not highlight groups at all? The current regex does not allow / to be part the username anyway, so we could as well check if the username is followed by a / in which case we don’t highlight it?

Sorry if this is obvious. But what's wrong with

```md
[@semantic-release](https://github.com/semantic-release)/generate-release-notes
```
?

gr2m:
@stevemao a text in the format of `@organization-name/other-name` is usually a team mention on GitHub, see https://github.com/blog/1121-introducing-team-mentions. But it’s can also be a scoped npm package name, see https://docs.npmjs.com/misc/scope

stevemao:
I see. We need a way to turn it off then. A follow up PR welcome :)

PR #221, opened 2017-09-07
docs: replace some old links with links to Lerna packages

aliceklipper:
Maybe badges in packages should be also updated? (With main repo's ones.)


stevemao:
Thanks!

PR #223, opened 2017-09-22
fix(lerna tags): Support multi-digit version tags

simondel:
We recently released `stryker@0.10.0` using Lerna. Ever since this module thinks that `stryker@0.9.3` is the last tag we have. This is because the minor and patch version are only allowed to be one digit. This PR fixes this issue by allowing multiple digits.

### DIFF
--- a/packages/git-semver-tags/index.js
+++ b/packages/git-semver-tags/index.js
@@ -8,7 +8,7 @@ function lernaTag(tag, pkg) {
   if (pkg && !(new RegExp('^' + pkg + '@')).test(tag)) {
     return false;
   } else {
-    return /^.+@[0-9]+\.[0-9]\.[0-9](-.+)?$/.test(tag);
+    return /^.+@[0-9]+\.[0-9]+\.[0-9]+(-.+)?$/.test(tag);
   }
 }
 
--- a/packages/git-semver-tags/test.js
+++ b/packages/git-semver-tags/test.js
@@ -128,6 +128,21 @@ it('should work with lerna style tags', function(done) {
   }, {lernaTags: true});
 });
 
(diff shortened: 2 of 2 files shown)


nicojs:
I feel like this cartoon is in order regex problem

evocateur:
Instead of string-parsing incredibly noisy output like `git log --decorate --no-color`, why not let git sort the tags ([properly](https://git-scm.com/docs/git-tag#git-tag---list), by version not lexically) for you? (not _you_, @simondel, sorry, I mean "this code, generally"):

```bash
# retrieve mixed annotated and lightweight tags
# annotated  : { "refname": "v2.0.0-beta.38", "tag": "245fbab", "commit": "7202ecc" }
# lightweight: { "refname": "package-name@1.0.1", "commit": "1b21c60", "": "" }
git tag --list --sort=version:refname --sort=creatordate --format='{ "refname": "%(refname:short)", "%(objecttype)": "%(objectname:short)", "%(*objecttype)": "%(*objectname:short)" }'

# the last item in the list returned will be the most recent tag
# you can even filter the pattern with fnmatch globs! e.g.:
#   independent lerna versions = `${CMD} '*@*'`
#   standard/"fixed" versions  = `${CMD} 'v*'`
```
([field name reference](https://git-scm.com/docs/git-for-each-ref#_field_names))

It would then be simpler to check the validity, split on `@` and pass the remainder to `semver.valid` (just like the non-lerna case) instead of maintaining a rickety regular expression.

simondel:
I favor a more sustainable solution, however I would appreciate if a fix could be released before that.

stevemao:
I will get onto this soon. Thanks for the patience.

PR #226, opened 2017-09-27
fix(filter): only remove commits that reverted commits in the scope

dlmr:
The change makes sure we only remove reverted commits if they reverted something in the current scope. 

This is useful when we are working with a subset of all commits and want to know if the commits makes sense to list in our changelog. Without this change we would always remove commits that made a revert from the changelog, even if the commit in question was done in a previous release. Now we will only remove the revert commit from the changelog if it reverted something in the current release.

Example using the Angular convention.

```
HASH-3 revert: fix: added something
                 This reverts commit HASH-1.
HASH-2 chore: release 1.0.1
HASH-1 fix: added something
```

This would make our changelog empty, since the filter with the default configuration would remove it. We would like it to list `revert: fix: added something`.

We would like it to remove our `revert:` commit however if the revert was done in the same "scope".

```
HASH-5: fix: solved that problem
HASH-4 revert: fix: added something else
                 This reverts commit HASH-3.
HASH-3 fix: added something else
HASH-2 chore: release 1.0.1
HASH-1 fix: added something
```

Here we expect that our changelog only have `fix: solved that problem` which this change still upholds.

### DIFF
--- a/packages/conventional-commits-filter/index.js
+++ b/packages/conventional-commits-filter/index.js
@@ -17,27 +17,39 @@ function conventionalCommitsFilter (commits) {
 
   var ret = [];
   var ignores = [];
+  var remove = [];
   commits.forEach(function (commit) {
     if (commit.revert) {
-      ignores.push(commit.revert);
-    } else {
-      ret.push(commit);
+      ignores.push(commit);
     }
+
+    ret.push(commit);
   });
 
+  // Filter out reverted commits
   ret = ret.filter(function (commit) {
     var ignoreThis = false;
 
     commit = commit.raw ? modifyValues(commit.raw, modifyValue) : modifyValues(commit, modifyValue);
 
-    ignores.some(function (ignore) {
-      ignore = modifyValues(ignore, modifyValue);
+    ignores.some(function (ignoreCommit) {
+      var ignore = modifyValues(ignoreCommit.revert, modifyValue);
 
       ignoreThis = isSubset(commit, ignore);
+
+      if (ignoreThis) {
+        remove.push(ignoreCommit.hash);
+      }
+
       return ignoreThis;
     });
 
     return !ignoreThis;
+  });
+
+  // Filter out the commits that reverted something otherwise keep the revert commits
+  ret = ret.filter(function (commit) {
+    return remove.indexOf(commit.hash) !== 0;
   });
 
   return ret;
--- a/packages/conventional-commits-filter/test.js
+++ b/packages/conventional-commits-filter/test.js
@@ -8,8 +8,39 @@ it('should error if `commits` is not `array`', function () {
   });
 });
 
-it('should filter reverted commits', function () {
+it('should filter reverted commits that exist in the commits array', function () {
   var commits = [{
+    type: 'revert',
+    scope: null,
+    subject: 'feat(): a very important feature',
+    header: 'revert: feat(): a very important feature\n',
+    body: 'This reverts commit 048fe156c9eddbe566f040f64ca6be1f55b16a23.\n',
+    footer: null,
+    notes: [],
+    references: [],
+    revert: {
+      header: 'feat(): amazing new module',
+      hash: '048fe156c9eddbe566f040f64ca6be1f55b16a23',
+      body: null
+    },
+    hash: '207abfa16885ef5ff88dfc6cdde694bb3fd03104\n',
+    raw: {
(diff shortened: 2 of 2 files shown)


dlmr:
@stevemao Is there anything I can do to help get this merged? 🙂

stevemao:
Sorry @dlmr for the delay. If I understand you correctly, Why don't you make two revert commits to revert two commits?

dlmr:
> Why don't you make two revert commits to revert two commits?

@stevemao Not sure I understand what you mean. We still need to add a `revert` for each commit we want to revert, this just makes the remove of the `revert` commit itself conditional. Before we always removed revert commits from the list of commits that we use for our changelog, now we **only** remove them **if** the revert commit reverted something in the current scope.

I tried to visualize this in my blocks above, that we want to "remove" the `revert` commit when we can remove the commit the `revert` commit reverts. If the commit that we are reverting already is something we have released we are not able to remove it from the changelog and therefor the revert action itself is something we should have in our changelog. With the current behaviour we will have an empty changelog if we revert something directly after a release when we actually would like to inform our users that the new release reverted something in the previous release.

I'm not sure I'm making my case any more clear but I think the tests and the examples should describe the situation in addition with this, if not I will try to explain again. 🙂

stevemao:
Because `HASH-1 fix: added something` was already released in `1.0.1`, and in the next release you want to revert this commit. But this fix was already released in `1.0.1`, and you can't really change the released app. Rather, if you do `revert: fix: added something` in the next release, and you want to inform your user that you've done something wrong in your current release.

If you change your released changelog, it might give your user false positive. I don't think you want to hide a commit from user, since the commit has already made into the app. The idea is similar with why npm forbids `npm publish -f`

dlmr:
I'm not talking about changing my released changelog, so I'm agreeing with you 100%. 😄 

Let me try to explain this better with a better(?) example.

**Git history before release**
```
21:00 HASH-3 revert: fix: added something
             This reverts commit HASH-1.
20:00 HASH-2 chore: release 1.0.1
19:00 HASH-1 fix: added something
...
```

We now create a new release, releasing `1.0.2` using `conventional-changelog`.

**Git history after release**
```
22:00 HASH-4 chore: release 1.0.2
21:00 HASH-3 revert: fix: added something
             This reverts commit HASH-1.
20:00 HASH-2 chore: release 1.0.1
19:00 HASH-1 fix: added something
...
```

Our git history looks correct, let's look at our `CHANGELOG.md`.

**Changelog with current behaviour**
```
# 1.0.2


# 1.0.1

### Fixes

* added something

## 1.0.0
```
The changes for 1.0.2 will be empty, this since this filter removes all commits that are revert commits and matches another commit, here the `This reverts commit [HASH]` is the key part.

**Changelog with PR behaviour**
```
## 1.0.2

### Reverts

* fix: added something


## 1.0.1

### Fixes

* added something

# 1.0.0
```
This is what I expect in the changelog, that we actually inform our users that something was reverted in our release.

dlmr:
> HASH-1 fix: added something should be in the 1.0.1 block, and HASH-6: revert: fix: added something, HASH-5: fix: solved that problem should be in the next release block.

@stevemao I agree with you fully and what you describe is what I want but that is NOT how the filter works today. This PR enables the behaviour you are describing. 

Today all revert commits will be removed so we will **only get HASH-5** in our changelog without this PR.

stevemao:
> but that is NOT how the filter works today.

hmmm... it's a bug then. I'll have a look. Thanks!

dlmr:
The problem from my understanding is this piece of code.

```js
if (commit.revert) {
  ignores.push(commit.revert);
} else {
  ret.push(commit);
}
```
https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-commits-filter/index.js#L21-L25

This means that we will "skip" all commits that are revert commits even if they don't revert anything in the current scope. What the PR does is to only skip/remove revert commits if they reverted something in the scope.

dlmr:
I hate to nag you about this but I would really like to see this merged. Please tell me if there is __anything__ I can do to help. 🙂

stevemao:
@dlmr I added you to the repo so you can merge it yourself.

dlmr:
I have tested this more in the time since our last conversation to verify that this works as expected and I can't find any indications that this change is incorrect and will therefore merge it, thanks for the help! 👍

PR #230, opened 2017-10-03
feat: allow forced cwd

marionebl:
This adds an optional second `opts` argument to `git-raw-commits` intended for options-passing to `child_process.execFile` - specifically this allows to force a cwd for the `git log` process.

**Edit**:
This allows consumers like [commitlint](https://github.com/marionebl/commitlint/blob/master/%40commitlint/core/src/read.js#L37) to decouple command from execution cwd. 

In the first step this makes testing considerably easier. Before I had to change the `cwd` via `process.chdir()` to pick up commit messages from prepared fixture repositories (which forced serial, slow execution) in my tests, now I can specify it via the new option.

For users, this is exposed up to the command line interface to allow for easier scripting, e.g.:

Before
```
cd git-repository
commitlint --from=HEAD^1
cd ..
``` 

After
```
commitlint --from=HEAD^1 --cwd=git-repository
```

### DIFF
--- a/packages/git-raw-commits/index.js
+++ b/packages/git-raw-commits/index.js
@@ -6,47 +6,60 @@ var stream = require('stream');
 var template = require('lodash.template');
 var through = require('through2');
 
-function gitRawCommits(options, opts) {
+var DELIMITER = '------------------------ >8 ------------------------';
+
+function normalizeExecOpts(execOpts) {
+  execOpts = execOpts || {};
+  execOpts.cwd = execOpts.cwd || process.cwd();
+  return execOpts;
+}
+
+function normalizeGitOpts(gitOpts) {
+  gitOpts = gitOpts || {};
+  gitOpts.format = gitOpts.format || '%B';
+  gitOpts.from = gitOpts.from || '';
+  gitOpts.to = gitOpts.to || 'HEAD';
+  return gitOpts;
+}
+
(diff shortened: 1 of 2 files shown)


stevemao:
Thanks @marionebl . I just need to know that how this is much better for you and what you have to do with the current behaviour. If you could write up more details it'd be awesome!

marionebl:
Hey @stevemao, sure - I edited the PR description accordingly. Ping me if you have more questions.

marionebl:
@stevemao Could you have another look?

packages/git-raw-commits/index.js
```diff
@@ -40,6 +41,7 @@ function gitRawCommits(options) {
   var isError = false;
 
   var child = execFile('git', args, {
```
stevemao: If you pass `options.cwd`, would it take it as an option already?

marionebl: I could pluck `cwd` from options, too. I thought it would be wise to keep git-specific options separated from ones geared towards the mode of execution / `process.execFile`

stevemao: Maybe rename it to something like `gitOpts` then :)

marionebl:
@stevemao: Can you publish a new release? I'd be happy to publish if you add me as maintainer with the `conventional-changelog` repos.

PR #231, opened 2017-10-04
fix(conventional-commits-parser): ignore comments 

marionebl:
* Ignore git comments
* Add relevant test cases


marionebl:
@stevemao Could you a look at this?

stevemao:
I don't think I will get you back soon... So I added you and you can merge yourself :)

stevemao:
I kinda feel there might be a better way to handle this though.

marionebl:
Thanks for adding me! I'd adapt the solution if you can pinpoint how the current one bothers you.

stevemao:
Actually, I'd probably add a `comment` section. But go ahead and do what you think is right. 😄

marionebl:
I'm going ahead with this. If we find use cases that need information about the comments I'll add it.

marionebl:
@stevemao: Can you publish a new release? I'd be happy to publish if you add me as maintainer with the `conventional-changelog` repos.

PR #237, opened 2017-10-16
fix(writer): normalize release headings (#214)

### LINKED ISSUE #214: why are patch version headings a different size?
Hello! 👋 

My apologies if this has already been brought up, but I wasn't able to find a similar question in past issues.

I noticed patch version headings are a different size from major and minor headings, which is a bit... unconventional? For example:

```
<a name="2.3.1"></a>
## [2.3.1](https://github.com/conventional-changelog/standard-version/compare/v2.3.0...v2.3.1) (2016-06-15)

...

<a name="2.3.0"></a>
# [2.3.0](https://github.com/conventional-changelog/standard-version/compare/v2.2.1...v2.3.0) (2016-06-02)
```

Inconsistent heading sizes break a changelog parser which I help maintain ([changelog-parser](https://github.com/hypermodules/changelog-parser)), which in turn is used by a github release tool I help maintain ([gh-release](https://github.com/hypermodules/gh-release)), which is used by a handful of projects that might benefit from using `standard-version` (roughly 500 projects). Right now the changelog convention we use is http://keepachangelog.com/, which `conventional-changelog` is *very* similar to.

The only real difference that makes parsing less straightforward is the different heading sizes.

It's also a bit problematic from an HTML/markdown perspective as h1 tags (`#`) are usually reserved for document titles and not typically used more than once in a single markdown file.

Is it in the realm of possibility for this project to adopt the convention of consistent version headings?

Thank you!

ungoldman:
Updated the default header template to generate semantic HTML headings using the method suggested by @marcysutton in #214 ✨ 

**Edit:** updated commit message with the following for context:

> Logic for generating release headings has been changed to make all
> heading levels the same (`##`/`h2`) for better compatibility with
> screen readers and parsers, and to conform to HTML semantics. Patch
> release titles are now wrapped in a `<small>` tag to maintain the
> visual hierarchy of the previous style.
> 
> Resolves #214.
> 
> Related issues:
> - #214 (comment)
> - conventional-changelog/standard-version#208

### DIFF
--- a/packages/conventional-changelog-core/test/test.js
+++ b/packages/conventional-changelog-core/test/test.js
@@ -250,7 +250,7 @@ describe('conventionalChangelogCore', function() {
       .pipe(through(function(chunk) {
         chunk = chunk.toString();
 
-        expect(chunk).to.include('## 0.0.17');
+        expect(chunk).to.include('## <small>0.0.17');
         expect(chunk).to.include('Second commit');
         expect(chunk).to.include('closes [#1](https://github.com/ajoslin/conventional-changelog/issues/1)');
 
@@ -269,7 +269,7 @@ describe('conventionalChangelogCore', function() {
       .pipe(through(function(chunk) {
         chunk = chunk.toString();
 
-        expect(chunk).to.include('## 0.0.17');
+        expect(chunk).to.include('## <small>0.0.17');
         expect(chunk).to.include('Second commit');
 
         done();
@@ -330,7 +330,7 @@ describe('conventionalChangelogCore', function() {
       .pipe(through(function(chunk) {
         chunk = chunk.toString();
 
-        expect(chunk).to.include('## v0.0.17');
+        expect(chunk).to.include('## <small>v0.0.17');
         expect(chunk).to.include('Second commit');
         expect(chunk).to.include('closes [#1](https://github.com/a/b/issues/1)');
 
@@ -367,7 +367,7 @@ describe('conventionalChangelogCore', function() {
     }).pipe(through(function(chunk) {
       chunk = chunk.toString();
 
-      expect(chunk).to.include('## 0.0.17');
+      expect(chunk).to.include('## <small>0.0.17');
       expect(chunk).to.include('closes [#1](github/a/b/issues/1)');
 
       done();
--- a/packages/conventional-changelog-writer/templates/header.hbs
+++ b/packages/conventional-changelog-writer/templates/header.hbs
@@ -1,11 +1,10 @@
 <a name="{{version}}"></a>
-{{#if isPatch~}}
-  ##
-{{~else~}}
-  #
-{{~/if}} {{version}}
+## {{#if isPatch~}} <small>
+  {{~/if~}} {{version}}
   {{~#if title}} "{{title}}"
-  {{~/if}}
+  {{~/if~}}
   {{~#if date}} ({{date}})
+  {{~/if~}}
+  {{~#if isPatch~}} </small>
   {{~/if}}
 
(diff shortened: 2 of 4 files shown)


e-cloud:
@ungoldman I think we could turn patch title into level 3. If each release's header is the same level(level2), it would be unconvinent for analysis purpose. It's better that we provide more semantic info.

stevemao:
Also CI must be green before merging, thanks!

hbetts:
@e-cloud in the template, @ungoldman is using `<small>`. Would that achieve the same desired effect?

@ungoldman we'd love to get this into `conventional-changelog`. Would you mind getting those CI tests green so we can accept your change?

Also, would you mind summarizing @marcysutton's comment in your commit message? That will make it much easier for us to understand these changes in the future when we're looking at commit annotations.

marcysutton:
`<small>` isn't the same as an `<h3>`....only the latter will communicate a different heading level. Small is for styling purposes only, so they aren't equivalent. My two cents :)

hbetts:
Thank you @marcysutton for the explanation.

I guess at this point I'm a little confused as to what we're trying to achieve here.

The pull request header says we're trying to generate semantic headers, yet the changes actually eliminate semantic differences between different release types by making every release a H2 header.

On the other hand this change does make headers consistent for the purpose of making it easier to parse changelogs (which I think was the point of #214)

So, @ungoldman and @marcysutton it would help if I knew what the goal of this change was, and if the pull request message was updated to reflect that purpose.

stevemao:
@hbetts https://github.com/conventional-changelog/conventional-changelog/issues/214#issuecomment-326681934 should answer your question. @ungoldman can you fix the tests and modify your commit/PR message that'd be great! Thanks!

ungoldman:
@hbetts Re: semantics, I think the inferred definition is being interpreted in more than one way -- the idea @marcysutton and I are suggesting is to make the headings semantic in terms of Markdown & HTML, so that there is a clear information hierarchy based on heading levels, e.g. `h2`/`##` consistently denotes a release, and anything below that denotes content related to that release.

```
h1 - title
├─ h2 - release
│  └─ h3 - features
│  └─ h3 - fixes
└─ h2 - release
```

This is beneficial not only for parsers, but also perhaps more importantly for [screen readers](https://en.wikipedia.org/wiki/Screen_reader).

I understand the desire to show a visual difference between major, minor, and patch releases, but I think that would be a difference in visual hierarchy, not semantic. Showing a stylistic difference between releases could be achieved with post-processing, or just the simple addition of a `<small>` tag, which was the compromise reached in https://github.com/conventional-changelog/conventional-changelog/issues/214#issuecomment-326681934 that this PR is trying to incorporate into the codebase.

@stevemao I'll make sure to devote some time to rebasing and getting tests passing ASAP.

Thanks all for your patience and encouragement.

hbetts:
>  think the inferred definition is being interpreted in more than one way 

@ungoldman thank you for explaining that the semantics are related to highlighting the differences between the _release_ name, and then parts of a release, such as _features_ or _fixes_.

I think it all makes sense to me now. Thank you very much! 😄

ungoldman:
I've updated the commit message with more contextual information and links to related issues, and got all tests passing locally. Waiting on Travis & Appveyor but this should be good to go pending maintainer approval.

@stevemao @hbetts does this constitute a breaking change?

stevemao:
Let's be a bit conservative and treat this as a breaking change.

hbetts:
> @stevemao @hbetts does this constitute a breaking change?

I would also side with this being a breaking change. To me it would seem the format of the changelog would be part of our API, and downstream consumers may be depending on that format.

packages/conventional-changelog-core/test/test.js
```diff
@@ -269,7 +269,7 @@ describe('conventionalChangelogCore', function() {
       .pipe(through(function(chunk) {
         chunk = chunk.toString();
 
-        expect(chunk).to.include('## 0.0.17');
+        expect(chunk).to.include('## <small>0.0.17');
```
hbetts: Is there anyway we can check that `</`small>` is also in the title line?

(Actually, it looks like you test this below for `conventional-changelog-writer`. So would it benefit us to verify the same behavior in `core`?)

PR #244, opened 2017-11-05
feat(dateformat): Change dateformat to yyyy-mm-dd HH:MM:ss

### LINKED ISSUE #243: conventional-changelog-cli: Override dateformat
So I started using the cli and want my preset to order by committer date. The problem is that conventional-changelog-core formats the date by default to yyyy-mm-dd and it's not overwritable by the cli. My idea is just to add time to the format by default.

I need the correct date format to sort by committerDate which now doesn't make much sense if you have multiple commits on the same day.

Is there a reason why you strip out the time?

wardpeet:
Fixes #243

### DIFF
--- a/packages/conventional-changelog-core/lib/merge-config.js
+++ b/packages/conventional-changelog-core/lib/merge-config.js
@@ -80,7 +80,7 @@ function mergeConfig(options, context, gitRawCommitsOpts, parserOpts, writerOpts
       }
 
       if (commit.committerDate) {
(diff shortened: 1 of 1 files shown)


hbetts:
I'm a little confused by this change since the issue, #243, is that the CLI doesn't allow you to override this value.

wardpeet:
I'm closing this one, I could just set date in writerOpts to get this one working as well so I don't need it in cli

Issue #189, opened 2017-05-22
Generate changelog from specific commit


lookapanda:
Hello,

is it somehow possible to generate a changelog starting from a specific commit?
So that it ignores everything BEFORE that commit?

stevemao:
You can config [gitRawCommitsOpts](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-core#gitrawcommitsopts) to just log the specific commits you want to include in the CHANGELOG

briandipalma:
Added a PR for this, `-f 3a423bc` but it takes a commit-ish so it could be a tag or whatever the user wishes to use.

hbetts:
@lookapanda, @briandipalma submitted #247 to address this issue, but the consensus from the core `conventional-changelog` group is to use a configuration file, and later, migrate the ecosystem to use `rc`-like files, as is being discussed in #195. 

@briandipalma has a good overview of their use case, and needs for `from` in the [following comment](https://github.com/conventional-changelog/conventional-changelog/pull/247#issuecomment-343887672).

Could the approach of using a configuration file, as mentioned by @briandipalma in #247, work for you as well? (At least until `rc` support is available?)

briandipalma:
We used a lerna style tagging approach which is described in our repo readme (https://github.com/caplin/caplin-dev-tools#releasing-packages) we then use the lerna cli flag `-l` and the created/updated changelog is fine for our purposes. So that might be another approach to configure a different `from` git commit-ish.

dangreen:
The CLI supports from/to with a tag or SHA after #1486. Closing as implemented.

Issue #195, opened 2017-06-05
get config from an rc file or package.json 


stevemao:
I've been looking at related projects such as commitizen and validate-commit-msg etc, and they all promote an rc file. I initially didn't want user to have too many dotfiles or rc files in their repo and worries this might discourage users to config. I was wrong. Seems like nobody knows how to config the tool.

stevemao:
CC @kentcdodds @jimthedev

aliceklipper:
I think we can use `cosmiconfig` to load configs from rc-files, `package.json` etc. E.g. from `conventional-changelog.config.js`, `conventional-changelogrc.yml` etc.

jimthedev:
For future reference @stevemao I don't think Kent uses commitizen / conventional-changelog anymore.

As far as commitizen goes, we're open to work on this a bit more with the community. Right now our rc files / package.json config loader was snagged from some other project. It looks something like: https://github.com/commitizen/cz-cli/blob/247d2c4a9756249c54d84471ba51b1c9ab6b9401/src/configLoader/loader.js

It is a bit of a mess and highly specific to us. One of our challenges is the complexity of our algorithm that we use to look for config. In commitizen land you can set a global convention for your machine that is used whenever a repo doesn't actually define its own convention.  This means looking for a local config first, then looking for a global config if none is found. It gets pretty hairy. I think currently allow for either a config in package.json or the rc file but the keys are specific to commitizen. We do not currently support .js configs or .yml config.

jimthedev:
Ok Here are a few more integration stories that show people who are either using or want to use commitizen (and by extension conventional-changelog) in interesting ways. I've resisted most major changes because the number one functionality when it comes to conventions (much like language specs) is maintain stability. It has worked well so far and considering I have limited bandwidth: laziness has been the (mostly) happy path.  Hopefully these issues give you a good idea of things that could be done in conjunction with conventional-changelog.

### User: I don't want a commitizen adapter as an npm dependency but do want commitizen to follow a specific conventional-changelog convention.

#### Description

Some people don't want to have a commitizen adapter as a dependency in their package (even though they are generally pretty small).  They want to just declare the adapter as a string in a config and then have commitzen cli download and use the adapter for their convention.  I've resisted on this one because I think npm does a better job of package management then anything I could write on my own. Sure declaring a convention/adapter is good, but really I want them to declare an implementation of that convention so that I can just use it.  

#### Possible solutions / paths forward

[...]

PR #247, opened 2017-11-10
feat(cli): Support specifing commit-ish to generate changelog from

briandipalma:
closes: #189

### DIFF
--- a/packages/conventional-changelog-cli/cli.js
+++ b/packages/conventional-changelog-cli/cli.js
@@ -47,7 +47,8 @@ var cli = meow({
     '',
     '  -c, --context             A filepath of a json that is used to define template variables',
     '  -l, --lerna-package       Generate a changelog for a specific lerna package (:pkg-name@1.0.0)',
-    '  --commit-path             Generate a changelog scoped to a specific directory'
+    '  --commit-path             Generate a changelog scoped to a specific directory',
+    '  -f --from                 Commit-ish to start generating changelog entries from'
   ]
 }, {
   alias: {
@@ -62,7 +63,8 @@ var cli = meow({
     v: 'verbose',
     n: 'config',
     c: 'context',
-    l: 'lernaPackage'
+    l: 'lernaPackage',
+    f: 'from'
   }
 });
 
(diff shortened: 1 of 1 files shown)


stevemao:
https://github.com/conventional-changelog/conventional-changelog/issues/195 should be a more generic solution.

briandipalma:
Hi @stevemao thanks for the reply. I'm not sure what the plan is with the rc file, there is a config file option but it seems underpowered. If a config file is supported why not allow it to configure everything e.g. preset, input/output, release-count, same-file etc? Currently the config file allows us to configure `from` and `path` of the git options but the other options have to be specified on the CLI.

The only option that's missing from the CLI for us is `from`. We are using the CLI in a monorepo with ~20 packages (https://github.com/caplin/caplin-dev-tools) so tagging the repo doesn't work as packages all have different versions. Tagging Lerna style might help but our repo structure is different. I'll give that a shot as we release packages and see if that helps us.

RomanMinkin:
@stevemao is there a way this PR can make it in since #195 is still in development? This one looks like a simple and elegant solution similar to `--release-count`.

briandipalma:
Happy to reopen if someone wants.

stevemao:
The problem with this PR is `from` is an option of `gitRawCommitsOpts`, not a conventional-changelog's `config`. Therefore, you'd add all options from the dependencies to the cli. 

If anyone want to create a PR to fix #195 that'd be great.

briandipalma:
I could change it to `from-sha` so it isn't the same?

RomanMinkin:
That's the beauty of `conventional-changelog-cli`, I just drop it into Makefile and forget, no config, no extra files.

In my case all I wanted is to strip all pre `1.0.0` text out of `CHANGELOG.md`. 
My current solution is to run [replace](https://www.npmjs.com/package/replace) right after the change log generation:
```bash
replace --silent '^(\#\s\[1\.0\.0\].+)([^]+)' '$1' CHANGELOG.md
```

stevemao:
If we want to add extra options in conventional-changelog, we should add it to conventional-changelog (or core), cli is just a proxy for the interface.

> That's the beauty of `conventional-changelog-cli`, I just drop it into Makefile and forget, no config, no extra files.

I believe that flags the same as config, or extra files. When it's small flags are nice. When it's big it becomes unmaintainable and better to put in extra files.

Most config in conventional-changelog should be either not a part of the submodules (context, gitRawCommitsOpts, parserOpts, writerOpts) to avoid duplicates, or to help coordinate across different submodules that are hard to config individually (`releaseCount` is a good example).

PR #253, opened 2017-11-28
fix: always parse references

### LINKED ISSUE #248: References are not picked up from header
## Actual

The code example below yields parser results without issues/references

## Expected

The code example below yields parser results with issues/references

```js
const parse = require('conventional-commits-parser').sync;
const angularPreset = require('conventional-changelog-angular');

(async () => {
  const preset = await angularPreset;
  parse('type: subject #1', preset.parserOpts);
  parse('type: subject\n #1', preset.parserOpts);
})();
```

## Related
* [marionebl/commitlint#115](https://github.com/marionebl/commitlint/issues/115)

marionebl:
* always parse references, regardless of presence of an action
* move reference parsing to reusable function
* adapt test cases to changed behaviour
* fixes #248

## Before

* references were dropped from header, body and footer by `parser.sync`
* this happened for references that had no `action` included in `options.referenceActions`
* due to `parser.sync` defaulting to an hardcoded set of `options.referenceActions`, references without (matching) action were never parsed

```js
 const parser = require('conventional-commits-parser').sync;

parse('type: subject #1').references;
/* [ ] */
```

## After
* fall back to catch-all regex if options.referenceActions regex produces no matches
* strip reference strings from all `notes`, was inconsistent before

```js
 const parser = require('conventional-commits-parser').sync;

parse('type: subject #1').references;
/* [ { action: null,
       owner: null,
       repository: null,
       issue: '1',
       raw: 'type: subject #1',
       prefix: '#' } ] */
```

### DIFF
--- a/packages/conventional-commits-parser/test/cli.spec.js
+++ b/packages/conventional-commits-parser/test/cli.spec.js
@@ -79,9 +79,44 @@ describe('cli', function() {
     });
     cp.stdout
       .pipe(concat(function(chunk) {
-        expect(chunk.toString()).to.include('"scope":"category","type":"fix:subcategory","subject":"My subject"');
-        expect(chunk.toString()).to.include('"references":[{"action":"Close","owner":null,"repository":null,"issue":"10036","raw":"#10036","prefix":"#"},{"action":"fix","owner":null,"repository":null,"issue":"9338","raw":"#9338","prefix":"#"}]');
-        expect(chunk.toString()).to.include('"notes":[{"title":"BREAKING NEWS","text":"A lot of changes!"}]');
+        var data = JSON.parse(chunk.toString())[0];
+        expect(data.scope).to.equal('category');
+        expect(data.type).to.equal('fix:subcategory');
+        expect(data.subject).to.equal('My subject');
+
+        expect(data.references).to.eql([
+          {
+            action: 'Close',
+            owner: null,
+            repository: null,
+            issue: '10036',
+            raw: '#10036',
+            prefix: '#'
+          },
+          {
+            action: null,
+            issue: '13233',
+            owner: null,
+            prefix: '#',
+            raw: 'Fixed #13233',
+            repository: null
+          },
+          {
+            action: 'fix',
+            owner: null,
+            repository: null,
+            issue: '9338',
+            raw: '#9338',
+            prefix: '#'
+          }
+        ]);
+
+        expect(data.notes).to.eql([
+          {
+            title: 'BREAKING NEWS',
+            text: 'A lot of changes!'
+          }
+        ]);
 
         done();
       }));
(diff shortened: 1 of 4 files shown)


packages/conventional-commits-parser/test/cli.spec.js
```diff
@@ -79,9 +79,44 @@ describe('cli', function() {
...
+        expect(data.references).to.eql([
+          {
+            action: 'Close',
+            owner: null,
+            repository: null,
+            issue: '10036',
+            raw: '#10036',
+            prefix: '#'
+          },
+          {
```
marionebl: This may be of interest, as it reflects the new behaviour for references not matched by the RegExp produced for `options.referenceActions`:

* `#13233` is parsed as reference
*  as `Fixed` is not in matched by `options.referenceActions`, action is `null`

packages/conventional-commits-parser/test/index.spec.js
```diff
@@ -170,6 +170,13 @@ describe('conventionalCommitsParser', function() {
             issue: '123',
             raw: '#123',
             prefix: '#'
+          }, {
```
marionebl: Reference previously was skipped due to non-matching action.

packages/conventional-commits-parser/test/index.spec.js
```diff
@@ -231,6 +239,13 @@ describe('conventionalCommitsParser', function() {
...
```
marionebl: Action previously was skpipped due to non-matching action

packages/conventional-commits-parser/test/index.spec.js
```diff
@@ -272,5 +287,45 @@ describe('sync', function() {
...
+        issue: '33',
+        owner: null,
+        prefix: '#',
+        raw: '#33',
+        repository: null
+      }
+    ]);
+  });
+
+  it('should parse references from header', function() {
```
marionebl: This is the minimal test case failing before this

packages/conventional-commits-parser/test/parser.spec.js
```diff
@@ -525,11 +525,12 @@ describe('parser', function() {
...
-      var text = 'Previously multiple template bindings on one element\n' +
-          '(ex. `<div *ngIf=\'..\' *ngFor=\'...\'>`) were allowed but most of the time\n' +
-          'were leading to undesired result. It is possible that a small number\n' +
-          'of applications will see template parse errors that shuld be fixed by\n' +
-          'nesting elements or using `<template>` tags explicitly.\n' +
+      var expectedText = 'Previously multiple template bindings on one element\n' +
+        '(ex. `<div *ngIf=\'..\' *ngFor=\'...\'>`) were allowed but most of the time\n' +
+        'were leading to undesired result. It is possible that a small number\n' +
+        'of applications will see template parse errors that shuld be fixed by\n' +
+        'nesting elements or using `<template>` tags explicitly.';
```
marionebl: The asterisk-extracted notes text omits the reference text now. This matches the behaviour of other notes-related tests.

marionebl:
Already did: https://github.com/conventional-changelog/conventional-changelog/pull/252

A review would be awesome!

stevemao:
@marionebl Feel free to merge this when you are ready! 😄

PR #254, opened 2017-11-29
fix(cli): set options.config to loaded custom config for processing

### LINKED ISSUE #227: BUG: Using a Custom Preset Config Not Working With `conventional-changelog-cli`
# Problem

@stevemao I created a custom preset by forking the conventional-changelog-angular. It can get it produce the expected output by dropping it into `/usr/local/lib/node_modules/conventional-changelog-cli/node_modules`. It is, however, not working with the --config option of the CLI, which is my preferred way of using it. It simply outputs the changelog without any expected sections, e.g. Added, Removed, etc.

# Steps to reproduce
 
1. Install the preset in an existing project with a few commits `npm install --save-dev conventional-changelog-stalinkay`
 2. Use `conventional-changelog --config ./node_modules/conventional-changelog-stalinkay/index.js -r 0`

# Expected Behavior

I expect changelog to contain the custom sections when using a custom --config but none are present when using the steps above.

RayBenefield:
Closes #227 

Setting this up allows the core to properly load the configuration (handling promises or however the config is delivered). Otherwise nothing is loaded and a default config is used instead.

### DIFF
--- a/packages/conventional-changelog-cli/cli.js
+++ b/packages/conventional-changelog-cli/cli.js
@@ -112,6 +112,7 @@ try {
 
   if (flags.config) {
     config = require(resolve(process.cwd(), flags.config));
+    options.config = config;
   } else {
     config = {};
   }
--- a/packages/conventional-changelog-cli/test/fixtures/promise-config.js
+++ b/packages/conventional-changelog-cli/test/fixtures/promise-config.js
@@ -1,0 +1,8 @@
+'use strict';
+var Q = require('q');
+
+module.exports = Q.resolve({
+  writerOpts: {
+    mainTemplate: '{{commitGroups.[0].commits.[0].type}}{{testContext}}template'
+  }
+});
--- a/packages/conventional-changelog-cli/package.json
+++ b/packages/conventional-changelog-cli/package.json
@@ -29,6 +29,7 @@
     "jscs": "^2.7.0",
     "jshint": "^2.8.0",
     "mocha": "*",
+    "q": "^1.5.1",
     "shelljs": "^0.6.0"
   },
   "dependencies": {
--- a/packages/conventional-changelog-cli/test/test.js
+++ b/packages/conventional-changelog-cli/test/test.js
@@ -312,6 +312,18 @@ describe('cli', function () {
       }));
   });
 
+  it('--config should work with a return promise', function (done) {
+    var cp = spawn(cliPath, ['--config', '../test/fixtures/promise-config.js'], {
+      stdio: [process.stdin, null, null]
+    });
+
+    cp.stdout
+      .pipe(concat(function (chunk) {
+        expect(chunk.toString()).to.equal('template');
+        done();
+      }));
+  });
+
   it('--config should work with relative path', function (done) {
     var cp = spawn(cliPath, ['--config', '../test/fixtures/config.js'], {
       stdio: [process.stdin, null, null]


RayBenefield:
Working on a test for this.

RayBenefield:
Alright test added. I had to add `q` to setup a proper promise.

hbetts:
Thank you for filing this @RayBenefield!

I hadn't even realized this wasn't getting passed through to `core`.

@stevemao does this change seem alright with you?

packages/conventional-changelog-cli/cli.js
```diff
@@ -112,6 +112,7 @@ try {
 
   if (flags.config) {
     config = require(resolve(process.cwd(), flags.config));
+    options.config = config;
```
hbetts: Could you just assign the required value to `options.config` and then use `options.config` elsewhere? That way there is only a single source of truth for the loaded configuration.

packages/conventional-changelog-cli/test/fixtures/promise-config.js
```diff
@@ -0,0 +1,8 @@
+'use strict';
+var Q = require('q');
```
hbetts: Is `q` needed? Could we just use native promises?

hbetts:
This might seem like an odd question, but if the configuration wasn't getting loaded, how do the other tests pass?

packages/conventional-changelog-cli/cli.js
RayBenefield: I'm not quite sure I follow. The main reason that I'm setting `options.config` here is because if config is a promise then `writerOpts`, `parserOpts`, and `gitRawCommitsOpts` won't be populated and can't be sent downstream. The other option is to resolve the promise in `conventional-changelog-cli` from the config, and then handle assignment of all of the opts after the resolution and then call `conventionalChangelog`.

The config gets resolved as a promise here: https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-core/lib/merge-config.js#L93

I modeled this change off of: https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog/index.js#L9

That is where presets are loaded to be resolved as a promise. And I don't see any other place that sets `options.config` so the only conflict that would appear would be if a custom config AND a preset was set. In this case, perhaps extending the custom config over top of the preset is something that should be done.

There are several ways to slice it and I'll leave it to you guys to decide as I'm not as familiar with the code base.

packages/conventional-changelog-cli/test/fixtures/promise-config.js
RayBenefield: I guess we could use native promises. I only used `q` to match the rest of the code base and I'm also not familiar with native promises availability in given versions of `node` so I didn't want to muddle any further. This is only like my third pull request and the others were simple readme changes. lol...

RayBenefield:
@hbetts the other tests are passing because when `writerOpts`, `parserOpts`, and `gitRawCommitsOpts` are extracted from `config` in the `cli` script, they exist when it isn't a promise, and don't exist when it is a promise. The resolution of a promisified config doesn't happen until it reaches `conventional-changlog-core`. Presets work fine because `options.config` is set after the `cli` script and is expected to be resolved in `conventional-changelog-core`.

packages/conventional-changelog-cli/test/fixtures/promise-config.js
hbetts: > This is only like my third pull request and the others were simple readme changes.

Ah, well thank you for taking the next step with this change.

> I only used q to match the rest of the code base 

That's fine. Since `conventional-changelog` only supports Node 4 and higher, so using native promises is perfectly fine, and my preference when possible (avoids third-party packages for native features where possible, lowering the overall overhead of installing `conventional-changelog`).

Though I'm not sure how much the other core maintainers of `conventional-changelog` prefer maintaining consistency within files.

packages/conventional-changelog-cli/cli.js
hbetts: Thank you for the explanation. I didn't realize you were relying on the promise resolution behavior of `conventional-changelog-core` to resolve `options.config`. This line makes a lot more sense now.

hbetts:
👍 This change looks fine to me.

machour:
Just tried `conventional-changelog-cli` for the first time and wanted to use my own templates .. thank you so much @RayBenefield for this PR, I was going nuts 😄

hbetts:
cc @stevemao and @bcoe for a last review before I accept this change.

hbetts:
@RayBenefield @machour, version `1.3.9` of `conventional-changelog-cli` has been released.

Successfully published the following packages:
 - conventional-changelog-angular@1.6.2
 - conventional-changelog-cli@1.3.9
 - conventional-changelog-core@2.0.1
 - conventional-changelog-ember@0.3.2
 - conventional-changelog@1.1.11
 - conventional-commits-parser@2.1.1
 - conventional-recommended-bump@1.2.1
 - gulp-conventional-changelog@1.1.11
 - standard-changelog@1.0.9

RayBenefield:
WOOT!!! First real PR to an external project. Thanks @hbetts for the experience. 👍

stevemao:
Sorry @FGRibreau, missed your issue https://github.com/conventional-changelog-archived-repos/conventional-changelog-cli/issues/9. it should be fixed by this. Thanks @RayBenefield !

Issue #241, opened 2017-10-24
Use Preset Packages


hbetts:
Re-posting an idea from conventional-changelog-archived-repos/conventional-recommended-bump/issues/3

I'm wondering if we can use the existing preset projects, such as `conventional-changelog-angular`, rather than needing to include the recommendation function for each preset as part of `conventional-recommended-bump`.

Doing that will necessitate using the API supported by `conventional-changelog`, which would make it easy for people to pass in custom presets.

As an example, `conventional-changelog-angular` returns back an object with `parserOpts` and `writerOpts`. `conventional-changelog-angular` could be extended to return a `bumpOpts` containing a `recommend` method. Then `conventional-recommended-bump` could take the preset module as an option under `options.config` similar to how `conventional-changelog` allows you to pass in custom presets.

hbetts:
The need has come up within `semantic-release-github` to support the [ability for users to pass in custom presets](https://gitlab.com/hyper-expanse/semantic-release-github/issues/3) to `conventional-changelog` to customize the changelog that's generated.

`semantic-release-github` already has a `--preset` CLI flag that allows users to specify a preset, and it applies to both how we determine the next version, and how we format the changelog.

However, the preset value is restricted to `angular` because [`conventional-recommended-bump`](https://www.npmjs.com/package/conventional-recommended-bump) only supports presets that are built into the package, which, at this time, only includes [`angular` and `eslint`](https://github.com/conventional-changelog/conventional-changelog/tree/c02b5b887c44ec055f4fe336e2c571fae3ecbd22/packages/conventional-recommended-bump/presets).

We can't allow any other values to be passed to the `--preset` CLI flag, including third-party `conventional-changelog` preset packages, because any other value for `--preset` would cause `conventional-recommended-bump` to fail.

Before I realized `semantic-release-github` already had a `--preset` CLI option, we planned on adding a `--changelog-preset` CLI flag for passing a preset package name to `conventional-changelog`. However, if a user wanted to set a preset they would be required to set the `--preset` CLI flag for the `conventional-recommended-bump` preset, and `--changelog-preset` CLI flag to set the preset for configuring the changelog output in `conventional-changelog`.

That's two preset flags.

Furthermore, not all `conventional-changelog` preset packages have an equivalent in `conventional-recommended-bump`.

Overall, it could become rather confusing and quite error prone.

hbetts:
My proposal is the following:

[...]

PR #256, opened 2017-12-01
feat(recommended-bump): add `eslint` preset

garyyeap:
As title, add eslint preset supports

### DIFF
--- a/packages/conventional-recommended-bump/presets/eslint.js
+++ b/packages/conventional-recommended-bump/presets/eslint.js
@@ -1,0 +1,39 @@
+var presetOpts = {
+  whatBump: function(commits) {
+    var level = 2;
+    var breakings = 0;
+    var features = 0;
+
+    commits.forEach(function(commit) {
+      if (!commit.type) {
+        return;
+      }
+
+      if (commit.type.toLowerCase() === 'breaking') {
+        breakings += 1;
+        level = 0;
+      } else if (commit.type.toLowerCase() === 'new') {
+        features += 1;
+        if (level === 2) {
+          level = 1;
+        }
+      }
+    });
+
+    return {
+      level: level,
+      reason: 'There are ' + breakings + ' breaking changes and ' + features + ' features'
+    };
+  },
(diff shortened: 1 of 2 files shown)


hbetts:
As I [commented here](https://github.com/conventional-changelog-archived-repos/conventional-recommended-bump/pull/23#issuecomment-339361967), I'm fine with accepting this change until #195 has been completed.

hbetts:
👍 looks good to me!

@stevemao?

stevemao:
Like I said, I'm not sure if anyone is using `preset` and I'm more towards user land `config` files.

garyyeap:
Thanks @hbetts, now I noticed the discussion on #195.
Since the content of `preset`(or `config`) for `recommended-bump` is totally different from `conventional-changelog` even is the same convention, I would like to propose an idea that we separate out `recommended-bump`'s presets like `conventional-changelog` does:

change the `require` action from

> config = require('./presets/' + options.preset);

to 

> options.config = require('conventional-recommended-bump-' + options.preset.toLowerCase());

as `conventional-changelog` does

> options.config = require('conventional-changelog-' + options.preset.toLowerCase());

Then I will close & change this PR to a third party package.

This may making more backward-compatibility when the final result of `rc` file discussion come out.

stevemao:
Yeah, they are configs for different projects. But the idea is the same. We should remove presets and do #195

Let's merge this first then we'll do #195 later :)

hbetts:
As a side note @stevemao and @garyyeap I kind of wonder if we can merge the two configurations together - #241

stevemao:
@hbetts That would be good. I also mentioned that if we could share config with tools like `commitizen` and `commitlint`.

garyyeap:
@hbetts First we have to catch up all existing `changelog` conventions for `recommended-bump`.
I think I can help with that and I'm wondering `conventional-changelog-*` presets still accepting PR?

hbetts:
@stevemao @garyyeap, if you don't mind, let's move our discussion to #241 😄

Issue #257, opened 2017-12-03
Alternate changelog links for gitlab?


flyinbutrs:
The compiled markdown file puts links to the commits in the changelog as:

````{{repo_url}}/commits/{{commit_hash}}````

However, gitlab uses: 

````{{repo_url}}/commit/{{commit_hash}}````

 Is there any ability to modify this?

**Comment by stevemao** [comment]
Is your host set to gitlab? https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-core/hosts/gitlab.json

**Comment by flyinbutrs** [comment]
No, how do I do that? I don't see it in the documentation anywhere.

**Comment by stevemao** [comment]
https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-core#host

**Comment by flyinbutrs** [comment]
Yeah, I found that in the docs, but I don't quite understand how to change that parameter. I assume by the docs there that it will default to the hostname of your repo, but in my case it's a private hosted gitlab, so it can't determine it from that hostname. Should I just add ```"host": "gitlab"``` to my ```package.json```?

**Comment by flyinbutrs** [comment]
Figured it out...  

in case anyone googles and is looking for the answer:

```"repository": "gitlab:<repo_url_here>"```

**Comment by flyinbutrs** [comment]
Actually, strike that...  I just noticed, when you do ```gitlab:<url>```, it assumes you want gitlab.com, and prefixes your repo URL with ```https://www.gitlab.com/```. Is there any way to use gitlab style issues/commit links but with a privately hosted gitlab?

**Comment by stevemao** [comment]
For now, you could overwrite any context value. 

```js
conventionalChangelogCore(options, {
    host: 'gitlab'
})
```

flyinbutrs:
Interesting. I'm trying to use standard-version / commitizen / standard-changelog with a variety of non-NPM packages (python, ruby, puppet), because I find it much friendlier out of the box than the other tools I've used for this type of build activity. 

That being said, I'm not really NPM "aware" in terms of what might be conventional in the JS world, and the docs are not really friendly to my use case (possibly for a reason). 

So...  where would I put that? Is there a guide I'm missing for how to "configure" this suite and where to put tweaks like this?

hbetts:
@flyinbutrs the sue of `standard-version` and `commitizen` is fine, and, by download count, used quite extensively within the JavaScript community.

[...]

PR #258, opened 2017-12-05
WIP: Allow repository.type to override autodetect type

flyinbutrs:
Related to #257, the current setup only allows the publicly hosted versions of gitlab/github/bitbucket. Unless I'm misreading the code, I think this can be fixed as follows:

I suspect this will allow you to do this syntax for privately hosted gitlab/github/bitbucket:

```
"repository": {
  "url": "https://private.gitlab",
  "type": "gitlab"
}
```

I know I need to add tests, and need to test this locally, but wanted to get confirmation first that I'm not barking up the wrong tree. I'm not a JS dev, so I'm probably making a lot of assumptions in how this works.

### DIFF
--- a/packages/conventional-changelog-core/lib/merge-config.js
+++ b/packages/conventional-changelog-core/lib/merge-config.js
@@ -198,7 +198,9 @@ function mergeConfig(options, context, gitRawCommitsOpts, parserOpts, writerOpts
       if (context.host && (!context.issue || !context.commit || !parserOpts || !parserOpts.referenceActions)) {
         var type;
 
-        if (context.host) {
+        if (context.type) {
+          type = context.type
+        } else if (context.host) {
           var match = context.host.match(rhosts);
           if (match) {
             type = match[0];
--- a/packages/conventional-changelog-core/lib/merge-config.js
+++ b/packages/conventional-changelog-core/lib/merge-config.js
@@ -199,7 +199,7 @@ function mergeConfig(options, context, gitRawCommitsOpts, parserOpts, writerOpts
         var type;
 
         if (context.type) {
-          type = context.type
+          type = context.type;
         } else if (context.host) {
           var match = context.host.match(rhosts);
           if (match) {


stevemao:
Sorry, I don't think this would be a scalable way to do it. There are many context values (you can even customise it). We can't provide all of them in this way. cosmiconfig should be more generic.

flyinbutrs:
Ok. I've read through a bunch of the packages involved in this (which you appear to be the primary maintainer of, so thanks!). Would it be possible to do it safely further up the chain?

If I"m reading this right, these are the current steps:

1. conventional-changelog-core loads package.json using read-pkg and defines it as pkg [[code](https://github.com/flyinbutrs/conventional-changelog/blob/f24b4753964c36b70b0287686aca8735941c44a5/packages/conventional-changelog-core/lib/merge-config.js#L103-L105)] 
1. if pkg or pkg.repository is not defined, get-remote-origin-url makes package.repository.url equal to the ```git remote origin -v```  [[code](https://github.com/flyinbutrs/conventional-changelog/blob/f24b4753964c36b70b0287686aca8735941c44a5/packages/conventional-changelog-core/lib/merge-config.js#L151-L154)]
1. repo is loaded using get-pkg-repo from the pkg object [[code](https://github.com/flyinbutrs/conventional-changelog/blob/f24b4753964c36b70b0287686aca8735941c44a5/packages/conventional-changelog-core/lib/merge-config.js#L161)]
    1. get-pkg-repo sets the type of the repo to either gitlab, github, or null based on the pkg.repository.url regex matching those domains or not. [[code](https://github.com/stevemao/get-pkg-repo/blob/45f78c7738fda6a662ecf5c1a7acc63c3ed9eada/index.js#L27-L34)]
1. repo.url is compared against the regex looking for `gitlab`, `github` or `bitbucket`. [[code](https://github.com/flyinbutrs/conventional-changelog/blob/f24b4753964c36b70b0287686aca8735941c44a5/packages/conventional-changelog-core/lib/merge-config.js#L204-L207)]
1. if repo.url does not match the above, and repo.type is defined, type is set to repo.type. [[code](https://github.com/flyinbutrs/conventional-changelog/blob/f24b4753964c36b70b0287686aca8735941c44a5/packages/conventional-changelog-core/lib/merge-config.js#L209)]

What I propose, instead of the above PR, changing [get-pkg-repo](https://github.com/stevemao/get-pkg-repo) so that unklownHostedInfo [[code](https://github.com/stevemao/get-pkg-repo/blob/ad3a6131b84fa5e85b1b5cbe4cfd0c4409c727a7/index.js#L36-L52)] so that if the [getType function](https://github.com/stevemao/get-pkg-repo/blob/ad3a6131b84fa5e85b1b5cbe4cfd0c4409c727a7/index.js#L27-L34) does not return a value, it will pass on a valid pkg.repository.type if it was set in package.json.

1. **CURRENT BEHAVIOR** - If repository.url regex matches a definitive domain via the getType function, repository.type is set to `gitlab`/`github` as the repo.type
1. **NEW BEHAVIOR** - If repository.url is not a regex match, and repository.type is defined as either `gitlab` or `github`, set repository.type accordingly
1. **CURRENT BEHAVIOR** - If neither of the above, set repository.type as `null`

PR #259, opened 2017-12-06
feat: allow to specify a tagPrefix in conventional-recommended-bump

nlunets:
This PR adds the possibility to specify a `tagPrefix` in `conventional-recommended-bump`.

Currently is you use a `tagPrefix` in https://github.com/conventional-changelog/standard-version and that tagPrefix is not semver compatible, standard version will not work as intended as it cannot find the releases.

By specifying the tagPrefix in `conventional-recommended-bump` and in turn in `git-semver-tags` we can get a proper version bump recommendation.

If this is accepted `standard-version` will have to be updated to pass the extra parameter to `conventional-recommended-bump`

Thanks in advance for your review

### DIFF
--- a/packages/conventional-recommended-bump/README.md
+++ b/packages/conventional-recommended-bump/README.md
@@ -77,6 +77,15 @@ An array of parsed commits. The commits are from last semver tag to `HEAD` and i
 
 If it returns with `level` `0` it will be a `major` bump. If `1`, `minor` bump. If `2`, `patch`.
 
+##### tagPrefix
+
+Type: `string`
+
+Specify a prefix for the git tag that will be taken into account during the comparison.
+
+For instance if your version tag is prefixed by `version/` instead of `v` you would specify `--tagPrefix=version/`
+
+
 #### parserOpts
 
 See the [conventional-commits-parser](https://github.com/conventional-changelog/conventional-commits-parser) docs.
@@ -96,12 +105,6 @@ Object includes what's returned by `whatBump` and
 Type: `string` Possible values: `'major'`, `'minor'` and `'patch'`
 
 The value of what it should release as.
-
-##### tagPrefix
-
-Type: `string`
-
-Specify a prefix for the git tag that will be taken into account during the comparison.
 
 ## Related
 
--- a/packages/conventional-recommended-bump/README.md
+++ b/packages/conventional-recommended-bump/README.md
@@ -97,6 +97,11 @@ Type: `string` Possible values: `'major'`, `'minor'` and `'patch'`
 
 The value of what it should release as.
 
+##### tagPrefix
+
+Type: `string`
+
+Specify a prefix for the git tag that will be taken into account during the comparison.
 
 ## Related
 
--- a/packages/git-semver-tags/cli.js
+++ b/packages/git-semver-tags/cli.js
@@ -10,7 +10,7 @@ var args = meow({
     'Options',
     ' --lerna parse lerna style git tags',
     ' --package when listing lerna style tags, filter by a package',
-    ' --tagPrefix prefix to remove from the tag during their processing'
+    ' --tagPrefix prefix to remove from the tags during their processing'
   ]
 });
 
--- a/packages/git-semver-tags/cli.js
+++ b/packages/git-semver-tags/cli.js
@@ -9,7 +9,8 @@ var args = meow({
     '  git-semver-tags',
     'Options',
     ' --lerna parse lerna style git tags',
-    ' --package when listing lerna style tags, filter by a package'
+    ' --package when listing lerna style tags, filter by a package',
+    ' --tagPrefix prefix to remove from the tag during their processing'
   ]
 });
 
@@ -22,5 +23,6 @@ gitSemverTags(function(err, tags) {
   console.log(tags.join('\n'));
 }, {
   lernaTags: args.flags.lerna,
-  package: args.flags.package
+  package: args.flags.package,
+  tagPrefix: args.flags.tagPrefix
 });
--- a/packages/git-semver-tags/index.js
+++ b/packages/git-semver-tags/index.js
@@ -29,7 +29,10 @@ module.exports = function(callback, opts) {
     }
 
     var tags = [];
-
+    var tagPrefixRegexp;
+    if (opts.tagPrefix) {
+      tagPrefixRegexp = new RegExp('^' + opts.tagPrefix);
+    }
     data.split('\n').forEach(function(decorations) {
       var match;
       while (match = regex.exec(decorations)) {
@@ -39,8 +42,8 @@ module.exports = function(callback, opts) {
             tags.push(tag);
           }
         } else if (opts.tagPrefix) {
-          if ((new RegExp('^' + opts.tagPrefix)).test(tag)) {
-            if (semverValid(tag.replace(new RegExp('^' + opts.tagPrefix), ''))) {
+          if (tagPrefixRegexp.test(tag)) {
+            if (semverValid(tag.replace(tagPrefixRegexp, ''))) {
               tags.push(tag);
             }
           }
--- a/packages/git-semver-tags/index.js
+++ b/packages/git-semver-tags/index.js
@@ -31,7 +31,7 @@ module.exports = function(callback, opts) {
     var tags = [];
     var tagPrefixRegexp;
     if (opts.tagPrefix) {
-      tagPrefixRegexp = new RegExp('^' + opts.tagPrefix);
+      tagPrefixRegexp = new RegExp('^' + opts.tagPrefix + '(.*)');
     }
     data.split('\n').forEach(function(decorations) {
       var match;
@@ -42,10 +42,9 @@ module.exports = function(callback, opts) {
             tags.push(tag);
           }
         } else if (opts.tagPrefix) {
-          if (tagPrefixRegexp.test(tag)) {
-            if (semverValid(tag.replace(tagPrefixRegexp, ''))) {
-              tags.push(tag);
-            }
+          var matches = tag.match(tagPrefixRegexp);
+          if (matches && semverValid(matches[1])) {
+            tags.push(tag);
           }
         } else if (semverValid(tag)) {
           tags.push(tag);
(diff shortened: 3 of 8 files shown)


hbetts:
I'll take a look at this later, but I believe this solves a need that recently came up in my own work. (Timing is amazing) Basically I need the ability to fetch tags with a prefix for the purpose of generating a recommended version bump for a single package within a monorepo.

nlunets:
@hbetts nice if it helps your use case ! If it doesn't completely fit your requirement let me know maybe I can adjust the code to be completely adequate.

packages/conventional-recommended-bump/README.md
```diff
@@ -97,6 +97,11 @@ Type: `string` Possible values: `'major'`, `'minor'` and `'patch'`
 
 The value of what it should release as.
 
+##### tagPrefix
```
hbetts: I think this section needs to be moved up, underneath the `#### options` section, instead of the `#### parserOpts` section. In the code below the option is pulled from `options` and not `parserOpts`.

packages/conventional-recommended-bump/README.md
```diff
@@ -97,6 +97,11 @@ Type: `string` Possible values: `'major'`, `'minor'` and `'patch'`
...
+
+Type: `string`
+
+Specify a prefix for the git tag that will be taken into account during the comparison.
```
hbetts: Could you provide an example of how this works in practice? Reading the code below, it seems obvious, but from just this one line of documentation, I'm not sure what this option does for me, or how I can use it.

packages/git-semver-tags/cli.js
```diff
@@ -9,7 +9,8 @@ var args = meow({
     '  git-semver-tags',
     'Options',
     ' --lerna parse lerna style git tags',
-    ' --package when listing lerna style tags, filter by a package'
+    ' --package when listing lerna style tags, filter by a package',
+    ' --tagPrefix prefix to remove from the tag during their processing'
```
hbetts: `tags during their`, or `tag during its`?

packages/git-semver-tags/index.js
```diff
@@ -38,6 +38,12 @@ module.exports = function(callback, opts) {
           if (lernaTag(tag, opts.package)) {
             tags.push(tag);
           }
+        } else if (opts.tagPrefix) {
+          if ((new RegExp('^' + opts.tagPrefix)).test(tag)) {
```
hbetts: Instead of constructing a regular expression on every tag, could this be moved outside of the `forEach` loop?

Also, does `tagPrefix` need to be escaped? It doesn't say anywhere that the `tagPrefix` can be a regular expressions. Therefore maybe it needs to be escaped before being passed to `RegExp`.

hbetts:
@nlunets actually I didn't notice we already had support for Lerna style tags, which I think is all I'll need for my particular use case.

Of course, that now has me wondering if your approach, supporting arbitrary tag prefixes, and the existing Lerna support, can be refactored so that we're not duplicating support here. Lerna style tags seem like nothing more than a special case of arbitrary tag prefixes.

hbetts:
Apparently the Lerna functionality isn't documented in the `README`. I also don't see where it was added in our commit history.

So I'm kind of in favor of making tag prefixes the standard and then documenting how to use the option for Lerna style stags.

cc @bcoe Lerna support first showed up when you imported `conventional-recommended-bump` into the monorepo. Therefore I can only assume it was added by you. Could I have your thoughts on generalizing Lerna support?

nlunets:
@hbetts I've adressed some of your comments, apart from the regexp escaping as I'm unsure what you'd like here :)

I also agree that lerna tags are a specific use case of this (although they don't check semverValid) but I'll let you and @bcoe evaluate what's the best way here. I'm fine with making changes based on your decision !

packages/conventional-recommended-bump/README.md
hellatan: isn't it preferable to style your tags with a suffix instead? something like `1.0.0-beta.1`, etc? By doing a suffix, it also seems to lend itself towards using npm's interface to increment the same release type:

```
command-line example:

$ semver 1.2.3 -i prerelease --preid beta
1.2.4-beta.0
Which then can be used to increment further:

$ semver 1.2.4-beta.0 -i prerelease
1.2.4-beta.1
```

https://docs.npmjs.com/misc/semver#prerelease-identifiers

nlunets: @hellatan sure, but sometimes you don't have a choice how your tags are built :) (especially in enterprise environment)

hbetts: > isn't it preferable to style your tags with a suffix instead? something like 1.0.0-beta.1, etc?

@hellatan, for the semantic version number, that is correct.

A prefix is not part of a valid semantic version (Though `v` was part of an earlier version of the [SemVer spec](https://semver.org/)).

However, prefixes are used quite often in the context of mono-repositories where individual packages within the mono-repository need to be tagged with version `1.0.0`. The only way to distinguish between the `1.0.0` tag for `package-a` versus `package-b` is to prefix the tag with the name of the package.

When `package-a` or `pakcage-b` is published, they are only published using the semantic version value of the tag, excluding the package prefix.

So if I were to increment the version for `package-a`, I would fetch all tags with the prefix `package-a`, remove the prefix, and then sort based on semantic versioning. Then I could pass the latest value to `semver` as you demonstrated.

hellatan: right...which we do at work for our mono-repo as well 😄. The project i'm working is a single repo so i'm not concerned with the prefix so it didn't really come to mind.

packages/git-semver-tags/test.js
```diff
@@ -160,3 +160,16 @@ it('should not allow package filter without lernaTags=true', function(done) {
     done();
   }, {package: 'bar-project'});
 });
+
+it('should work with tag prefix option', function(done) {
+  writeFileSync('test6', '');
+  shell.exec('git add --all && git commit -m"eigth commit"');
+  shell.exec('git tag ms/6.0.0');
+  shell.exec('git add --all && git commit -m"tenth commit"');
+  shell.exec('git tag ms/7.0.0');
```
hbetts: Can you also please add a tag where `ms/` is somewhere else in the tag.

The purpose is to ensure the code above is properly checking for the prefix at the beginning of the tag string.

packages/git-semver-tags/index.js
```diff
@@ -38,6 +41,12 @@ module.exports = function(callback, opts) {
           if (lernaTag(tag, opts.package)) {
             tags.push(tag);
           }
+        } else if (opts.tagPrefix) {
+          if (tagPrefixRegexp.test(tag)) {
+            if (semverValid(tag.replace(tagPrefixRegexp, ''))) {
```
hbetts: Could you use `String.match()` with a modified regex to reduce these two conditions down to one?

For example:

```javascript
const matches = tag.match(new RegExp(`^${opts.tagPrefix}(.*)`));
if (matches && semverValid(matches[0])) {}
```

hbetts:
@nlunets I provided a little bit more feedback.

Personally I think it may be best to accept this change, and then follow back around with a refactor that replaces the `lerna` feature, either completely, or by having it re-use the `tagPrefix` code.

nlunets:
@hbetts I've addressed your follow up comments (although dropping the es6 syntax as the linter does not accept it).
How do we move forward ?

nlunets:
Any update on a potential review for this @hbetts @bcoe now that the christmas break is over ? :)

hbetts:
I'm sorry @nlunets. I kind of lost track of this issue. Thank you for pinging me again.

👍 This work looks good to me.

@bcoe or @stevemao would it be possible to receive a second opinion to make sure I haven't missed anything?

stevemao:
Sorry for the late reply.

@hbetts `lernaTags` was added when we needed to support lerna in `conventional-changelog`

> Lerna style tags seem like nothing more than a special case of arbitrary tag prefixes.

What Lerna style tags can't do but tag prefixes can (let's just focus on `conventional-recommended-bump`)?

nlunets:
@stevemao lerna style tags assume that the format will be `package`@`version` which is ok when using lerna but which might not be the case for custom tags, as @hbetts stated it might be better in the long run to have lerna tags only at a higher level and rely on a generic functionality here to deal with both lerna tags and custom tags.

stevemao:
Hmmm... In that case, why not just change the regex https://github.com/conventional-changelog/conventional-changelog/pull/259/files#diff-879f31377f7a1c060d1596ecf0e410baR8
and maybe instead of calling it lerna, give it a more generic name?

nlunets:
As was discussed with @hbetts the goal was not to disrupt the current functionality.
`tagPrefix` is exposed in `standard-version` but is unusable because of the lack of support downstream today.

If you feel we need to merge the two functionality already then that's fine and I'll accomodate that.

hbetts:
> and maybe instead of calling it lerna, give it a more generic name?

@stevemao as mentioned by @nlunets, `the goal was not to disrupt the current functionality.`, by supporting both `lernaTags` and `tagPrefix` side-by-side.

Under the hood, in the future, yes, we should definitely have the Lerna function call the tag prefix function to avoid the tag detection duplication.

Now, whether we get rid of the `lernaTags` all together seems outside the scope of this particular change, and should be evaluated based on whether we think the extra flag (and internal function) provides more convenience to users than the cost to us to maintain.

stevemao:
@hbetts Sounds good :)

hbetts:
Successfully published the following from this pull request:
 - conventional-recommended-bump@1.2.0
 - git-semver-tags@1.3.0

Also successfully published:
 - conventional-changelog-angular@1.6.1
 - conventional-changelog-cli@1.3.8
 - conventional-changelog-core@2.0.0
 - conventional-changelog-ember@0.3.1
 - conventional-changelog-eslint@1.0.0
 - conventional-changelog-writer@3.0.0
 - conventional-changelog@1.1.10
 - gulp-conventional-changelog@1.1.10
 - standard-changelog@1.0.8

evocateur:
I'm totally gonna make lerna use `tagPrefix` now, this is sweet!

PR #262, opened 2017-12-10
feat(preset): add recommended-bump opts into presets

hbetts:
Move `conventional-recommended-bump` options from the built-in
presets into the individual `conventional-changelog-*` presets
so that the same preset packages may be used between
`conventional-recommended-bump` and `conventional-changelog`.

Related to #241


hbetts:
This change is ready to go.

stevemao:
I think this should be written in a different module/file. Would be easy to enable treeshaking. Or people can just require a single file as they need.

destroyerofbuilds:
> I think this should be written in a different module/file.

@stevemao, if I understand you correctly, I should move the `recommendedBumpOpts` object into a separate file, perhaps called `recommended-bump-options.js`, and then update the `index.js` file to effectively re-export the contents of `recommended-bump-options` through the same `recommendedBumpOpts` option?

That seems pretty easy and straight forward. 👍 

If I'm understanding what you're looking for, I'll make that change. 😄

stevemao:
Yeah. My original thought was to let `recommended-bump` (or maybe even `commitizen`) somehow read `changelog-config`'s config without changing too much `changelog-config`. I haven't had time to look into this. But your approach seems also good since we could install less npm packages. However if we separate the files/modules people can still have the same granularity and easier to refactor in the future if we ever want to publish separately. With the ES import/export treeshaking would also work fine.

I'm completely open to suggestions and would be great if other contributors can comment :)

hbetts:
Based on our conversation so far, I wanted to put together my thoughts on three different approach to this refactor just to make sure I'm on the same page with everyone else.

This is basically me just rambling on, trying to hit the high points of each possible approach. Some details are wrong (such as package names), so please keep that in mind.

---

**(1)** So to one extreme we could create a package per convention for each of the end-user tools we support (Such as `commitizen`, `conventional-changelog`, and `conventional-recommended-bump`), plus one that combines them all together.

In this scenario we would end up with three tool-specific packages for the Angular convention:
* `conventional-changelog-angular` - For `conventional-changelog`.
* `commitizen-angular` - For `commitizen`.
* `conventional-recommended-bump-angular` - For `conventional-recommended-bump`.

We would then re-export all these via `conventional-angular`.

That would be four packages for Angular.

With these packages, users of the end-user CLI tools could do:
```bash
conventional-recommended-bump --preset conventional-recommended-bump-angular
// or, using the meta package:
conventional-recommended-bump --preset conventional-angular
```

Benefits include the ability to use the configuration package for the end-user tool someone is using, plus tree-shaking.

Furthermore, for higher-level CLI tools, such as `semantic-release-github`, which use the `conventional-recommended-bump` and `conventional-changelog` packages, users could pass in the `conventional-angular` meta package (which can then be passed directly to the underlying tools' API).

A major drawback of this approach is the greater number of individual packages that need to be maintained within this monorepo.

**(2)** On the other extreme is the approach I took here, which is to combine all configuration into a single object, within a single file, within a single package.

With these packages, users of the end-user CLI tools could do:
```bash
conventional-recommended-bump --preset conventional-changelog-angular
```

Drawbacks include no tree-shaking support, and no ability to install, and use, the absolute minimum configuration needed for a single end-user CLI tool.

Benefits include a single configuration file, and single package, per convention that we need to maintain.

**(3)** As a balance between both approaches, and based on @stevemao's feedback, we could maintain tool-specific files within a single convention package.

For the Angular convention, we could maintain the `conventional-changelog-angular` package, but within that package, there would be the following:
```
/conventional-recommended-bump.js
/commitizen.js
/conventional-changlog.js
```

Furthermore, when configuration needs to be shared between export points, we could split that out as well.

For example, `parserOpts` needs to be shared between `conventional-changelog` and `conventional-recommended-bump`.

Therefore, we could split `parserOpts` into a `/parseropts.js` file and re-export that file's export through both `/conventional-changelog.js` and `/conventional-recommended-bump.js`.

`/index.js` would be a re-export of the three main files above.

The benefit is that we still maintain a single package per convention, but we also support tree-shaking through individual configuration files that can be imported.

In addition the configuration can be passed to end-user tools like so:
```bash
conventional-recommended-bump --preset conventional-changelog-angular/conventional-recommended-bump
// or, using the meta package:
conventional-recommended-bump --preset conventional-changelog-angular
```

A drawback is that to use any of the configuration, the entire package needs to be installed, so it's not possible to install just the configuration you need through a tool-specific package.

---

So I'm going to refactor my work to follow the third approach. (Please give me a day or two to swing back around and do that refactor)

stevemao:
Thanks @hbetts for the detailed writeup!

I'd like to hear opinions from other contributors as well especially @jimthedev :)

If we go for the third option then https://github.com/commitizen/cz-conventional-changelog will be deprecated. Looks like it's a bit out-dated anyway. Putting them in one place might help to keep it up to date.

hbetts:
I've refactored `conventional-changelog-angular` to match the third option.

As a side note, I updated the code to use ES6 features available in Node 4. Looks like support for Node 0.10 and 0.12 were removed back in March, so hopefully this isn't a breaking change on its own.

With the refactor there are dedicated `conventional-changelog` and `conventional-recommended-bump` files. I'm not sure how I feel about tool specific configuration files in `conventional-changelog-angular`, but it definitely helps in terms of maintenance, as all configuration is contained within one package.

Also note that this change exports `conventional-changelog` and `conventional-recommended-bump` along side `writer-opts` and `parser-opts`.

Perhaps when I merge `parser-opts`, I can remove the `parser-opts` and `writer-opts` exports.

hbetts:
This refactor is now done.

It seems I can merge the pull request, but I'll wait another 48 hours for any additional feedback before accepting.

Then I'll kindly reach out to someone to release the changes.

stevemao:
I have added you to the org on npm so you should be able to push. There are some packages that can't be added to the org and I already contacted npm support

PR #265, opened 2017-12-14
[WIP] feat: indicate patch/fix output for bump

hellatan:
it seemed weird/confusing to me that only features were indicated that something was bumped from the output message. Since this is supposed to follow semver, I think it makes sense to give some output for any sort of fix as well. thoughts?

### DIFF
--- a/packages/conventional-recommended-bump/presets/angular.js
+++ b/packages/conventional-recommended-bump/presets/angular.js
@@ -3,6 +3,7 @@ var presetOpts = {
     var level = 2;
     var breakings = 0;
     var features = 0;
+    var fixes = 0;
 
     commits.forEach(function(commit) {
       if (commit.notes.length > 0) {
@@ -13,12 +14,22 @@ var presetOpts = {
         if (level === 2) {
           level = 1;
         }
+      } else if (commit.type === 'fix') {
+        // no need to reset `level` since it is
+        // inherently a patch and it would be
+        // preferable to take the highest level bump
+        fixes += 1;
       }
     });
 
     return {
       level: level,
-      reason: 'There are ' + breakings + ' BREAKING CHANGES and ' + features + ' features'
+      reason: 'There are ' + breakings + ' BREAKING CHANGES, ' + features + ' features, and ' + fixes + ' fixes',
+      changes: {
+        breaking: breakings,
+        features: features,
+        fixes: fixes
+      }
     };
   },
   parserOpts: {
--- a/packages/conventional-recommended-bump/test/angular.js
+++ b/packages/conventional-recommended-bump/test/angular.js
@@ -18,6 +18,10 @@ betterThanBefore.setups([
     shell.exec('git add --all && git commit -m "feat($compile): new feature"');
     writeFileSync('test3', '');
     shell.exec('git add --all && git commit -m "perf(ngOptions): make it faster"');
+    writeFileSync('test3.1', '');
+    shell.exec('git add --all && git commit -m "fix: typo"');
+    writeFileSync('test3.2', '');
+    shell.exec('git add --all && git commit -m "fix: flip-flopped condition"');
   },
   function() { // 2
     writeFileSync('test4', '');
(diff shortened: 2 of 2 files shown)


stevemao:
The only types that matter is `feat` and `BREAKING CHANGE`. Even if you don't have any new commits, it'll be a patch. It doesn't matter if your commits have `chore`, `fix` or `ci` etc.

hellatan:
> The only types that matter is feat and BREAKING CHANGE

why is this the case? A patch shouldn't be conflated with a feature

hellatan:
oooh wait. i think i see what you're saying @stevemao. Are you saying you'd prefer not to have this change or to change `} else if (commit.type === 'fix') {` to just `} else {`?

stevemao:
That's what's happening now. If you don't have anything other than `feat` or `BREAKING CHANGE` it'll be a patch.

hellatan:
@stevemao are you for or against having the number of patches output? I think it's nice to indicate that there were some changes. Also, for consumers, having the `changes` object is nice to know as well.

stevemao:
Then you'd need to count `ci`, `build`, chore` and `doc` etc...

hellatan:
then could it be limited to `fix` just for the count? That's what seems to be what conventional commits goes by. I think it's confusing when you're expecting some patch output but then you actually don't get any indication that there may have been a patch since this current output only takes majors and minors into account.

hbetts:
Personally I would like more granularity as to the reasoning behind why `conventional-recommended-bump` suggested one type of bump over another. I would find it useful to know that it's releasing a patch because there are fixes to be released.

However, I'm not sure it's particularly useful here. 

As noted by @stevemao, the `angular` preset will _always_ recommend at _least_ a patch.

Even if there are no new commits, it will still recommend a patch. In that particular scenario, this particular change would not effectively enhance the user experience, because if there are no commits, the `reason` line would show `0` fixes. The user would continue to be confused ("Why did this just recommend a patch if there are no fixes?")

Also, why only show fixes? A `perf` commit, one that improves performance of an existing feature, may be a reason to release a `patch`, yet it won't show up in the `reason` message.

As an alternative to improve clarity, how about changing the `reason` line so that it will either show the existing message if there are breaking changes or enhancements, or, if there are none of those, have it say: "There are no breaking changes or enhancements, but this preset will always recommend at least a patch release".

stevemao:
> There are no breaking changes or enhancements, but this preset will always recommend at least a patch release

I love it!

hbetts:
@hellatan what do you think about changing the messaging as I suggested above?

hellatan:
sorry @hbetts, i'm on sabbatical right now so I'm slow to answer and remember things 😄. I'll make sure to take a closer look this coming week though and get back to you on it...being on the road can kind of get in the way of programming sometimes

hellatan:
@hbetts just to be clear, change this:
```js
reason: 'There are ' + breakings + ' BREAKING CHANGES, ' + features + ' features, and ' + fixes + ' fixes',
```

to be something more like this:

```js
var reasonMsg;
if (breakings === 0 && features === 0 && fixes === 0) {
    reasonMsg = "There are no breaking changes or enhancements, but this preset will always recommend at least a patch release";
} else {
    reason = 'There are ' + breakings + ' BREAKING CHANGES, ' + features + ' features, and ' + fixes + ' fixes';
}

return {
    reason: reasonMsg,
    // etc
};
```

this may me being anal retentive, but do you want me to change the plurality if there is only one change for any of the types so it reads more properly (ie. `1 feature` instead of `1 features`)?

hbetts:
> do you want me to change the plurality if there is only one change for any of the types so it reads more properly (ie. 1 feature instead of 1 features)?

Unless you want to, don't worry about it. The plurality of the message was already a part of the code, and though it would be _nice_ to have it fixed some day, it's not related to your work here.

> to be something more like this:

Correct. Also, I'm fine with the addition of `fixes`.

So overall, your example looks spot on with what I was thinking.

hellatan:
@hbetts great! I'll update accordingly

hellatan:
@hbetts @stevemao sorry, i really slacked on this and am catching up on the new file structure. I believe I want to move these changes here now instead - https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-angular/conventional-recommended-bump.js - is that correct?

hbetts:
> I believe I want to move these changes here now instead

That is correct. We made a pretty big change to how our presets are written, and how they are used by `conventional-recommended-bump`.

hbetts:
I am closing this pull request for now. Please feel free to re-submit the pull request if you have time to make progress on this work.

hellatan:
@hbetts will do. the timing of my changes and the timing of your migration earlier this year really put a 🔧 in this PR 😄

PR #267, opened 2017-12-26
fix: truncate after scissors line

marionebl:
This resolves an issue where commit authored via `git commit -v` and fed into `conventional-commits-parser` via `.git/COMMIT_EDITMSG` directly where not truncated as intended from the beginning of the scissors line

## Before

https://runkit.com/marionebl/conventional-commits-parser-scissors
```js 
const parser = require('conventional-commits-parser');
const message = `A message
# ------------------------ >8 ------------------------
diff --git a/package.json b/package.json
index d88860d05..dd6c7818f 100644
--- a/somehting.json
+++ b/somehting.json`;

parser.sync(message, {commentChar: '#'}) // => {body: 'diff --git a/package.json b/packag…mehting.json\n+++ b/somehting.json'}
```

## After

```
const parser = require('conventional-commits-parser');
const message = `A message
# ------------------------ >8 ------------------------
diff --git a/package.json b/package.json
index d88860d05..dd6c7818f 100644
--- a/somehting.json
+++ b/somehting.json`;

parser.sync(message, {commentChar: '#'}) // => {body: null}
```

### DIFF
--- a/packages/conventional-commits-parser/lib/parser.js
+++ b/packages/conventional-commits-parser/lib/parser.js
@@ -21,6 +21,16 @@ function getCommentFilter(char) {
   };
 }
 
+function truncateToScissor(lines) {
+  var scissorIndex = lines.indexOf(SCISSOR);
+
+  if (scissorIndex === -1) {
+    return lines;
+  }
+
+  return lines.slice(0, scissorIndex);
+}
+
 function getReferences(input, regex) {
   var references = [];
   var referenceSentences;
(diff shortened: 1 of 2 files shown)


packages/conventional-commits-parser/lib/parser.js
```diff
@@ -86,7 +87,13 @@ function parser(raw, options, regex) {
     getCommentFilter(options.commentChar) :
     passTrough;
 
-  var lines = trimOffNewlines(raw).split(/\r?\n/).filter(commentFilter);
+  var rawLines = trimOffNewlines(raw).split(/\r?\n/);
+  var scissorIndex = rawLines.indexOf(SCISSOR);
```
bcoe: I'd be tempted to pull these few lines into a helper method; perhaps `truncateToScissor`?

packages/conventional-commits-parser/test/parser.spec.js
```diff
@@ -285,6 +285,17 @@ describe('parser', function() {
     });
   });
 
+  it('should truncate from scissors line', function() {
+    var msg = parser(
+      'this is some subject before a scissors-line\n' +
```
bcoe: I'd be tempted to assert that the subject line has been populated appropriately (ensuring that your truncation logic wasn't too aggressive).

marionebl: Opted for additional cases here to describe explicity what was tested.

marionebl:
@bcoe Could you have another look?

PR #268, opened 2018-01-08
feat: regex headerPattern for eslint

JPeer264:
The previous regex got a lot of overhead. The new one is easier to read and faster to parse.

But here I changed the regex a bit. First it also matched whitespaces in the beginning of the message, now it doesnt. If the same behavior is wished, I can change the regex to:

```js
 /^(\w*): (.*)$/
```
Which does the same as the original one, but with less overhead

### DIFF
--- a/packages/conventional-changelog-eslint/parser-opts.js
+++ b/packages/conventional-changelog-eslint/parser-opts.js
@@ -1,7 +1,7 @@
 'use strict';
 
 module.exports = {
-  headerPattern: /^(\w*)\: (.*?)(?:\((.*)\))?$/,
+  headerPattern: /^(\w*):\s*(.*)$/,
   headerCorrespondence: [
     `tag`,
     `message`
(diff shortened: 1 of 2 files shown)


stevemao:
Can you add a test?

JPeer264:
I added a test case with the beginning trailing whitespaces. Did you mean this test case?

hbetts:
👍 I'm fine with this change.

I will note that I couldn't actually find any ESLint documentation on whether commit messages are allowed to contain spaces after the commit type and colon.

Their [documention](https://eslint.org/docs/developer-guide/contributing/pull-requests#step-2-make-your-changes) does say one way or the other.

@stevemao should this be a breaking change since the whitespace behavior has changed?

JPeer264:
I would see this as breaking change, since this feature won't apply to the previous style. Also a fix wouldn't fit, 'cause the lib was never thought to be that way.

stevemao:
@JPeer264 ok. Thanks! Could you squash and change the commit message to follow our standard? Make sure you have a `BREAKING CHANGE:` in it.

JPeer264:
like this?

PR #270, opened 2018-01-16
feat: support `conventional-changelog-*` presets in `conventional-recommended-bump`

hbetts:
Add support to pass in a `conventional-changelog-*` preset package
name through a `preset` option in the options argument passed to
`conventional-recommended-bump`.

Previous behavior of the `preset` option was to map the name provided
through the option to one of the presets built into this
`conventional-recommended-bump` package.

Now we take the name passed to the `preset` option, whether passed
programmatically or via the CLI, pass it to
`conventional-changelog-preset-loader`, and use the returned result as
the preset package.

The loading behavior matches the behavior of the `preset` option in
`conventional-changelog`. Including support for the different export
types of conventional preset packages, such as a promise, function,
or object.

Lastly, we defer to `recommendedBumpOpts.parserOpts` property of a preset
package if it exists, as our initial refactor efforts created a
`parserOpts` object under `recommendedBumpOpts` in each preset package.

In the future we will merge differences found in
`recommendedBumpOpts.parserOpts` into the top-level `parserOpts` object
of each `conventional-changelog-*` preset and remove
`recommendedBumpOpts.parserOpts` from each preset package, if it exists.

BREAKING CHANGES:

First:

Will only accept the name associated with a `conventional-changelog-*`
preset package for the purpose of loading preset configuration.

Built-in presets previously available in `conventional-recommended-bump`
are no longer available.

The preset package name can be any value supported by
[`conventional-changelog-preset-loader`](https://www.npmjs.com/package/conventional-changelog-preset-loader).

Related to #241

Second:

Remove backwards-compatible support for `whatBump` to return a number.

If `whatBump` returns a number it will no longer be mapped to an object
with a `level` and `releaseType` property. Please update the passed in
`whatBump` function to return an object with a `level` property
indicating the type of bump (`0` for `major, `1` for `minor`, and `2`
for `patch`).

Third:

A callback function must be provided to `conventional-recommended-bump`,
either in the second argument spot, in place of `parserOpts`, or in the
third argument spot, technically reserved for a callback.


hbetts:
This work is related to #241.

Specifically bullet points [3, 4, and 5](https://github.com/conventional-changelog/conventional-changelog/issues/241#issuecomment-350317418)

This pull request is a _Work in Progress_ until I can write tests that account for the new behavior of `conventional-recommended-bump`.

packages/conventional-recommended-bump/index.js
```diff
@@ -4,86 +4,88 @@ var conventionalCommitsFilter = require('conventional-commits-filter');
...
-    ignoreReverted: true,
-    warn: function() {}
-  }, options);
+  cb = typeof parserOpts === 'function' ? cb = parserOpts : cb = cb || noop;
 
+  let presetPackage;
   if (options.preset) {
     try {
-      config = require('./presets/' + options.preset);
+      presetPackage = require(`conventional-changelog-${options.preset.toLowerCase()}`);
```
stevemao: What do you think of removing preset option and adding a `config.extends` (similar to what eslint does). It accepts a string (or an array of strings) of either relative or absolute path to a module.

fenduru: This will prevent scope packages from working

stevemao: @fenduru this would simplify scope packages, since a path is more generic.

hbetts: > What do you think of removing preset option and adding a config.extends (similar to what eslint does). It accepts a string (or an array of strings) of either relative or absolute path to a module.

@stevemao we could do that. Though that would mean `conventional-changelog` and `conventional-recommended-bump` would remain different in how they load presets (unless we rolled `conventional-changelog` into this PR.)

hbetts: > This will prevent scope packages from working

My current implementation does not support npm scopes (`@my-org`). That is correct.

My hope is to borrow the implementation from #272, once complete.

> adding a config.extends

@stevemao if you're interested in that kind of change to the API, maybe we should roll it into #272 where we're changing the behavior of presets to add scope support.

It should be noted that ESLint [behaves](https://eslint.org/docs/developer-guide/shareable-configs#using-a-shareable-config) similar to the work in #272 _plus_ support for arbitrary paths, though ESLint requires that you specify the full package name when using scopes, instead of assuming the package name like Yeoman, Yarn, and #272 do.

hbetts:
@stevemao @bcoe @fenduru this Pull Request is ready for final review before I merge it into the `master` branch.

packages/conventional-recommended-bump/cli.js
```diff
@@ -1,55 +1,57 @@
...
-    '  -c, --header-correspondence    Comma separated parts used to define what capturing group of `headerPattern` captures what',
-    '  -r, --reference-actions        Comma separated keywords that used to reference issues',
-    '  -i, --issue-prefixes           Comma separated prefixes of an issue',
-    '  -n, --note-keywords            Comma separated keywords for important notes',
-    '  -f, --field-pattern            Regex to match other fields',
-    '  -v, --verbose                  Verbose output',
-    '  -l, --lerna-package            Recommend a bump for a specific lerna package (:pkg-name@1.0.0)',
-    '  -t, --tag-prefix               Tag prefix to consider when reading the tags',
-    '  --commit-path                  Recommend a bump scoped to a specific directory'
+    `Usage`,
```
stevemao: with template literals, we don't need to use an array.

hbetts: Fixed.

packages/conventional-recommended-bump/cli.js
```diff
@@ -1,55 +1,54 @@
...
-    '  -l, --lerna-package            Recommend a bump for a specific lerna package (:pkg-name@1.0.0)',
-    '  -t, --tag-prefix               Tag prefix to consider when reading the tags',
-    '  --commit-path                  Recommend a bump scoped to a specific directory'
-  ]
-}, {
+const meow = require(`meow`);
+const conventionalRecommendedBump = require(`./`);
+const path = require(`path`);
+
+const cli = meow(`
```
stevemao: I think there will be an extra newline and indentations here in the help menu.

hbetts: According to [`meow`](https://www.npmjs.com/package/meow#help) - `The input is reindented and starting/ending newlines are trimmed which means you can use a template literal without having to care about using the correct amount of indent.`

On the command line it looks like:

```bash
(17:59 CST)-(feat/presets)--> ./cli.js --help

  Get a recommended version bump based on conventional commits

  Usage
    conventional-recommended-bump

  Example
    conventional-recommended-bump

  Options
    -p, --preset                   Name of the preset you want to use
    -g, --config                   A filepath of your config script
    -h, --header-pattern           Regex to match header pattern
    -c, --header-correspondence    Comma separated parts used to define what capturing group of 'headerPattern' captures what
    -r, --reference-actions        Comma separated keywords that used to reference issues
    -i, --issue-prefixes           Comma separated prefixes of an issue
    -n, --note-keywords            Comma separated keywords for important notes
    -f, --field-pattern            Regex to match other fields
    -v, --verbose                  Verbose output
    -l, --lerna-package            Recommend a bump for a specific lerna package (:pkg-name@1.0.0)
    -t, --tag-prefix               Tag prefix to consider when reading the tags
    --commit-path                  Recommend a bump scoped to a specific directory

```

hbetts:
Version [2.0.0](https://github.com/conventional-changelog/conventional-changelog/releases/tag/conventional-recommended-bump%402.0.0) of `conventional-recommended-bump` has been published.

PR #272, opened 2018-01-23
fix(presets): support scoped packages as presets

fenduru:
Previously, a scoped packages such as `@my-org/conventional-changelog-myorg` was not able to be used when specifying a preset. This change checks if the preset is scoped, and re-appends the scope when constructing the preset package name.

### DIFF
--- a/packages/conventional-changelog/index.js
+++ b/packages/conventional-changelog/index.js
@@ -15,7 +15,6 @@ function conventionalChangelog(options, context, gitRawCommitsOpts, parserOpts, 
         name = parts.join('');
       }
       
-      console.log('scope', scope, 'name', name);
       options.config = require(scope + 'conventional-changelog-' + name);
     } catch (err) {
       options.warn('Preset: "' + options.preset + '" does not exist');
--- a/packages/conventional-changelog/index.js
+++ b/packages/conventional-changelog/index.js
@@ -14,7 +14,7 @@ function conventionalChangelog(options, context, gitRawCommitsOpts, parserOpts, 
         scope = parts.shift() + '/';
         name = parts.join('');
       }
-      
+
       options.config = require(scope + 'conventional-changelog-' + name);
     } catch (err) {
       options.warn('Preset: "' + options.preset + '" does not exist');
--- a/packages/conventional-changelog/index.js
+++ b/packages/conventional-changelog/index.js
@@ -6,7 +6,17 @@ function conventionalChangelog(options, context, gitRawCommitsOpts, parserOpts, 
 
   if (options.preset) {
     try {
-      options.config = require('conventional-changelog-' + options.preset.toLowerCase());
+      var scope = '';
+      var name = options.preset.toLowerCase();
+
+      if (name[0] === '@') {
+        var parts = name.split('/');
+        scope = parts.shift() + '/';
+        name = parts.join('');
+      }
+      
+      console.log('scope', scope, 'name', name);
+      options.config = require(scope + 'conventional-changelog-' + name);
     } catch (err) {
       options.warn('Preset: "' + options.preset + '" does not exist');
     }
--- a/packages/conventional-changelog/README.md
+++ b/packages/conventional-changelog/README.md
@@ -48,6 +48,8 @@ Type: `string` Possible values: `'angular', 'atom', 'codemirror', 'ember', 'esli
 
 It's recommended to use a preset so you don't have to define everything yourself. Presets are names of built-in `config`.
 
+A scoped preset package such as `@scope/conventional-changelog-custom-preset` can be used by passing `@scope/custom-preset` to this option.
+
 **NOTE:** `options.config` will be overwritten by the values of preset. You should use either `preset` or `config`, but not both.
 
 


packages/conventional-changelog/index.js
```diff
@@ -6,7 +6,16 @@ function conventionalChangelog(options, context, gitRawCommitsOpts, parserOpts,
 
   if (options.preset) {
     try {
-      options.config = require('conventional-changelog-' + options.preset.toLowerCase());
+      var scope = '';
+      var name = options.preset.toLowerCase();
```
hbetts: There could be upper-case preset packages (At one point `npm` allowed upper case package names, so `npm info JSON` is not the same as `npm info json`).

I would just respect the casing provided by the user.

packages/conventional-changelog/index.js
```diff
@@ -6,7 +6,16 @@ function conventionalChangelog(options, context, gitRawCommitsOpts, parserOpts,
...
+
+      if (name[0] === '@') {
+        var parts = name.split('/');
+        scope = parts.shift() + '/';
+        name = parts.join('');
```
hbetts: There will never be anything here to join. A forward slash may only be used to delineate between a scope and package name. [Here's the relevant validation code](https://github.com/npm/validate-npm-package-name/blob/393f65720d07eca49e9f1672f0cf001b71b021d9/index.js#L77).

hbetts:
@fenduru do you mind updating the [`conventional-changelog` documentation](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog#preset) to demonstrate the new scope support?

hbetts:
Also, would you mind adding a test to verify proper scope behavior? (Relevant test section - https://github.com/conventional-changelog/conventional-changelog/blob/9e87dc32b4ffc5202661e674f20e5901feb1ac6f/packages/conventional-changelog/test/test.js#L24)

packages/conventional-changelog/index.js
fenduru: The `toLowerCase` is existing code so that seems a bit out of scope for this PR

packages/conventional-changelog/index.js
fenduru: `npm` may validate package names like this, but I'm not sure that this validation is done in `conventional-changelog`. The package is simply required, so if someone was previously passing `--preset my-preset/lib/file.js` that would translate to `require('conventional-changelog-my-preset/lib/file.js')` which is valid.

I think this is needed to avoid accidental breakage

fenduru:
I'll add to the docs and tests tomorrow 👍

packages/conventional-changelog/index.js
hbetts: > The toLowerCase is existing code so that seems a bit out of scope for this PR

That's a good point. It would be out of scope.

packages/conventional-changelog/index.js
hbetts: Thank you for catching that.

hbetts:
So on the prompting of @stevemao I looked at ESLint's [config sharing](https://eslint.org/docs/developer-guide/shareable-configs#using-a-shareable-config) rules and they specifically require that you include the full package name when using scopes.

Their reasoning:

> Doing so would result in package naming conflicts, and thus in resolution errors in most of cases. For example a package named @scope/eslint-config-myconfig vs @scope/my-config, since both are valid scoped package names

I don't know how that's not the case with unscoped packages.

fenduru:
@hbetts I don't see how there is a conflict there. Supporting full package names is obviously more flexible (i.e. I could use a package that doesn't start with `conventional-commit-`), but is different than the current API.

I'm going to leave it as is for now due to consistency with what `--preset` currently supports. @stevemao if you feel strongly one way or the other let me know and I'll change it if necessary.

hbetts:
Thank you @fenduru!

This looks good to me other than needing a test and some documentation.

packages/conventional-changelog/index.js
```diff
@@ -6,7 +6,16 @@ function conventionalChangelog(options, context, gitRawCommitsOpts, parserOpts,
 
   if (options.preset) {
     try {
-      options.config = require('conventional-changelog-' + options.preset.toLowerCase());
+      var scope = '';
+      var name = options.preset.toLowerCase();
+
+      if (name[0] === '@') {
```
stevemao: I'm pretty sure we can use an npm package to achieve this.

hbetts: There is [`is-scoped`](https://www.npmjs.com/package/is-scoped), but it wont separate the package name into it's parts.

fenduru: There is [this packages](https://github.com/keithamus/parse-packagejson-name). Seems like overkill to me though

hbetts: Nor does that package look supported.

Personally, except for `is-scoped`, which looks supported, I wouldn't go with a package that isn't receiving at least some attention once in a while

`is-scoped` probably isn't worth it here either since we already have to do package name manipulation. We don't really help ourselves avoid any complexity.

fenduru:
I added a line to the docs, but I'm not really sure how to test this (I can't tell what is going on in the existing tests).

hbetts:
@fenduru the unit tests for `conventional-changelog`, such as:

```javascript
conventionalChangelog({
      preset: 'angular',
      warn: function(warning) {
        done(warning);
      }
})
```

... rely on the fact that `conventional-changelog-angular` is listed as a runtime dependency of the `conventional-changelog` package, and that the unit tests run as a sub-directory within the `conventional-changelog` folder. Therefore, when the `conventional-changelog` code `require`'s the `angular` preset, the `require` goes up two directories, from the temporary test directory, to the `node_modules/` directory within `conventional-changelog`, and loads the preset.

This is to say, the fastest way to verify scope support is to list a scoped preset package as a dependency of `conventional-changelog`, and require it in a copy/paste of an existing unit test statement.

A slightly more lengthy, but I believe more correct, way is to create a mock `node_modules/` directory containing a mocked scope preset package, and write the unit test to require that.

fenduru:
> A slightly more lengthy, but I believe more correct, way is to create a mock node_modules/ directory containing a mocked scope preset package, and write the unit test to require that.

I'm not sure how I should do that.

It generally seems weird that unit tests rely on package.json dependencies as well as filesystem.

packages/conventional-changelog/index.js
hbetts: @evocateur what about the following scenario:

You list the preset as `@mycompany/angular` with the expectation that it will load the `@mycompany/conventional-changelog-angular` package listed in your development dependencies. One of your dependencies lists `@mycompany/angular` as a dependency, which is installed at the top level `node_modules/` directory because `npm` and `yarn` both flatten the `node_modules/` directory as best they can.

When `conventional-changelog` goes to `require` the `preset` you listed, which is `@mycompany/angular`, it would try loading the actual `@mycompany/angular` package, not the `@mycompany/conventional-changelog-angular` package you intended.

evocateur:
Use the explicit package name, not the shorthand? Simple enough to recognize a scoped attempt at implicit resolution and fail with a warning, akin to the eslint approach (which I tend to prefer, anyhow).

hbetts:
I've released version `1.1.0` of `conventional-changelog-preset-loader` (which is a sub-package of this mono-repository) - https://www.npmjs.com/package/conventional-changelog-preset-loader

It contains the loader logic implemented by @fenduru, and unit tests that don't rely on file system interactions, or nesting of directory structure.

It's already been introduced into the `conventional-recommended-bump` refactor in #270.

@evocateur if there is a name conflict, the solution is, as you noted, to use the full preset package name. However, because a naming conflict can arise at anytime, I believe you either need to work with only `conventional-changelog-*` preset package names, or you need to only support explicit package names. Perhaps you can check for naming conflicts and fail, but the act of introducing a naming conflict would be a breaking change for the project using a preset package, which wouldn't be captured by semver in the dependency that introduced the conflicting package name (because they wouldn't know it was a conflict).

It would love to see us come to a consensus on whether we should be using only explicit names, or supporting the shorthand approach extended in this PR. Would you, @evocateur, consider filing an issue to start the discussion on which strategy we should use going forward for `conventional-changelog` packages?

hbetts:
Version [`1.1.12`](https://github.com/conventional-changelog/conventional-changelog/releases/tag/conventional-changelog%401.1.12) of `conventional-changelog` has been released with this fix.

PR #275, opened 2018-02-01
test(core): create isolated test directory

hbetts:


hbetts:
@stevemao as noted by @TheDancingCode in #274, the act of releasing new versions of `conventional-changelog` actually broke the tests.

Though I don't know _how_ the new releases broke the tests, the tests were breaking because they were reading git tags from this project from within the temporary test directories created by the unit tests.

Therefore, I created temporary test directories outside of the project folder and that seems to have alleviated the issue.

PR #278, opened 2018-02-12
fix(preset-loader): don't namespace exported function

hbetts:


hbetts:
cc @evocateur 

Looks like my original code for the preset loader added an extra namespace, which wasn't intended, and wasn't documented in the `README`. Therefore I'm removing the extra namespace.

evocateur:
Looks good!

hbetts:
Version [1.1.2](https://github.com/conventional-changelog/conventional-changelog/releases/tag/conventional-changelog-preset-loader%401.1.2) has been released.

PR #281, opened 2018-02-13
fix: roll back conventional-changelog-preset-loader

rarkins:


hbetts:
@bcoe has successfully published version `1.1.3` of `conventional-changelog-preset-loader`.

This change is no longer necessary.

PR #293, opened 2018-02-28
Switch to ESLint

### LINKED ISSUE #228: Use eslint instead of jscs/jshint
We should remove individual linting configs/deps per package and just create a "global" one

TheDancingCode:
Implements ESLint with `standard` code style.

Closes #228.


stevemao:
@TheDancingCode thanks! Would it be easier just to use `standard` module?

TheDancingCode:
Yes, but I understood that @hbetts [prefers it this way](https://github.com/conventional-changelog/conventional-changelog/issues/228#issuecomment-368356130):

> My only opinion in this area is that we use the `standard` ESLint configuration package, instead of the `standard` package itself. I believe more developers will have their IDEs setup with an ESLint plugin than a plugin for `standard`.

hbetts:
> Would it be easier just to use standard module?

@stevemao as @TheDancingCode mentioned, I was the one that recommended using ESLint and the `standard` plugin/config.

As an example, I personally use Visual Studio Code as my primary editor, and the ESLint plugin, which I have installed, has about 5 million downloads, while the Standard plugin has bout 100 thousand. That implies to me a greater likelihood that contributors to this project would be setup with ESLint, and would not need to undertake additional setup steps. It also implies a greater mind share around ESLint that will make it easier to on-board future contributors.

package.json
```diff
@@ -34,10 +34,14 @@
   },
   "version": "1.0.0",
   "devDependencies": {
-    "coveralls": "^3.0.0",
     "chai": "^4.1.2",
-    "jscs": "^3.0.7",
-    "jshint": "^2.9.5",
+    "coveralls": "^3.0.0",
+    "eslint": "4.18.1",
```
hbetts: Can these please be updated to use the `^` notation? That way it's implied that we expect our project to work with all future `patch` and `minor` releases.

packages/conventional-changelog-angular/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "conventional-changelog angular preset",
   "main": "index.js",
   "scripts": {
-    "lint": "jshint *.js --exclude node_modules && jscs *.js",
+    "lint": "eslint --fix .",
```
hbetts: I hadn't thought of enabling the auto-fix option (I don't do that in my own personal projects), but I guess I'm fine with it.

packages/conventional-changelog-angular/test/test.js
```diff
@@ -1,334 +1,337 @@
...
-var conventionalChangelogCore = require('conventional-changelog-core');
-var preset = require('../');
-var expect = require('chai').expect;
-var gitDummyCommit = require('git-dummy-commit');
-var shell = require('shelljs');
-var through = require('through2');
-var betterThanBefore = require('better-than-before')();
-var preparing = betterThanBefore.preparing;
+'use strict'
+var conventionalChangelogCore = require('conventional-changelog-core')
```
hbetts: I'm a little disappointed `standard` doesn't encourage the use of `const`, or even outright ban the use of `var`. Oh well. I'll probably switch these all over later.

packages/conventional-changelog-writer/test/partial.commit.spec.js
```diff
@@ -1,19 +1,25 @@
...
+var describe = mocha.describe
+var fs = require('fs')
+var Handlebars = require('handlebars')
+
+var template
+var templateContext
+
+before(function (done) {
+  fs.readFile('templates/commit.hbs', function (err, data) {
+    if (err) done(err)
```
hbetts: Nice! Glad to see this standard already helping to catch uncaught errors.

hbetts:
Except for my comment regarding the semantic version range in `package.json`, this change looks good to me :+1:

package.json
TheDancingCode: I will do that. Forgot to undo some recent npm configuration.

packages/conventional-changelog-angular/test/test.js
TheDancingCode: If need be, I can add the [no-var](https://eslint.org/docs/rules/no-var) rule to the config.

TheDancingCode:
Code is now updated according to the requested changes.

packages/conventional-changelog-angular/test/test.js
hbetts: No, let's wait :smile: 

This change is to just adopt the `standard` rules for now, and then we can discuss enforcing `const` across all the `conventional-changelog` related repositories.

hbetts:
@TheDancingCode would you please rebase your pull request?

I needed to accept #295, which has led to a conflict in `conventional-changelog-eslint`.

TheDancingCode:
Rebase done.

PR #296, opened 2018-03-19
[Angular Package] Adapt to new Angular Commit Conventions

flxwu:
Adapted to the new Angular Commit Conventions according to https://github.com/angular/angular/blob/master/CONTRIBUTING.md#type

### DIFF
--- a/packages/conventional-changelog-angular/conventional-recommended-bump.js
+++ b/packages/conventional-changelog-angular/conventional-recommended-bump.js
@@ -20,7 +20,9 @@ module.exports = {
 
     return {
       level: level,
-      reason: `There are ${breakings} BREAKING CHANGES and ${features} features`
+      reason: breakings === 1 ?
+        `There are ${breakings} BREAKING CHANGE and ${features} features` :
+        `There are ${breakings} BREAKING CHANGES and ${features} features`
     }
   },
 
--- a/packages/conventional-changelog-angular/test/test.js
+++ b/packages/conventional-changelog-angular/test/test.js
@@ -22,7 +22,7 @@ betterThanBefore.setups([
     shell.exec('git init --template=./git-templates')
 
     gitDummyCommit(['build: first build setup', 'BREAKING CHANGE: New build system.'])
-    gitDummyCommit(['ci(circle): add CircleCI pipeline', 'BREAKING CHANGE: Continuously integrated.'])
+    gitDummyCommit(['ci(travis): add TravisCI pipeline', 'BREAKING CHANGE: Continuously integrated.'])
     gitDummyCommit(['feat: amazing new module', 'BREAKING CHANGE: Not backward compatible.'])
     gitDummyCommit(['fix(compile): avoid a bug', 'BREAKING CHANGE: The Change is huge.'])
     gitDummyCommit(['perf(ngOptions): make it faster', ' closes #1, #2'])
@@ -72,7 +72,8 @@ describe('angular preset', function () {
         chunk = chunk.toString()
 
         expect(chunk).to.include('first build setup')
-        expect(chunk).to.include('**circle:** Continuously integrated.')
+        expect(chunk).to.include('**travis:** add TravisCI pipeline')
+        expect(chunk).to.include('**travis:** Continuously integrated.')
         expect(chunk).to.include('amazing new module')
         expect(chunk).to.include('**compile:** avoid a bug')
         expect(chunk).to.include('make it faster')
@@ -89,7 +90,6 @@ describe('angular preset', function () {
         expect(chunk).to.include('bad commit')
         expect(chunk).to.include('BREAKING CHANGE')
 
-        expect(chunk).to.not.include('first build setup')
         expect(chunk).to.not.include('ci')
         expect(chunk).to.not.include('feat')
         expect(chunk).to.not.include('fix')
@@ -175,7 +175,7 @@ describe('angular preset', function () {
   })
 
   it('should work if there is a semver tag', function (done) {
-    preparing(7)
+    preparing(6)
     var i = 0
 
     conventionalChangelogCore({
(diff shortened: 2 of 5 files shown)


packages/conventional-changelog-angular/writer-opts.js
```diff
@@ -51,8 +51,10 @@ function getWriterOpts () {
         commit.type = `Code Refactoring`
       } else if (commit.type === `test`) {
         commit.type = `Tests`
-      } else if (commit.type === `chore`) {
-        commit.type = `Chores`
+      } else if (commit.type === `build`) {
+        commit.type = `Build System`
+      } else if (commit.type === `ci`) {
+        commit.type = `Continious Integration`
```
alfaproject: `Continuous`

flxwu: Sorry :relieved:

hbetts:
@flxwu thank you for submitting an update to the Angular conventions. It's been long overdue.

Before we accept this change, can you please update your commit messages to conform to our [Conventional- Commits](https://conventionalcommits.org/) pattern? In particular, this will need to be marked as a breaking change since it replaces what commit types are supported.

flxwu:
Yeah sure, I was also about to add a mesage to the cli tool that outputs a note regarding this change @hbetts

hbetts:
On a related note, the Angular conventions specifically say that breaking changes must start with `BREAKING CHANGE`, yet the Angular convention supports the plural form of `CHANGE` (`CHANGES`).

To be in compliance with the Angular convention, I think we need to also remove the plural form of `BREAKING CHANGE`.

flxwu:
Can I just output to console in the angular index.js? @hbetts

flxwu:
Just reworded to match the commit conventions

packages/conventional-changelog-angular/conventional-recommended-bump.js
```diff
@@ -20,7 +20,9 @@ module.exports = {
 
     return {
       level: level,
-      reason: `There are ${breakings} BREAKING CHANGES and ${features} features`
+      reason: breakings === 1 ?
+        `There are ${breakings} BREAKING CHANGE and ${features} features` :
```
hbetts: `There are` => `There is`

hbetts:
> Can I just output to console in the angular index.js?

Kind of depends. 

What are you hoping to achieve by printing something to console?

packages/conventional-changelog-angular/conventional-recommended-bump.js
flxwu: But There are Change and features....

hbetts:
@flxwu can you please update the unit tests to support your changes?

For example, this [`chore`](https://github.com/conventional-changelog/conventional-changelog/blob/d146bdc0e64d90bd04dc361e8602576bc95f36dc/packages/conventional-changelog-angular/test/test.js#L45) commit will need to be replaced.

flxwu:
To inform the user about the new naming convention

hbetts:
> To inform the user about the new naming convention

To adopt the new convention users would need to upgrade to a new major version (Since this is a breaking change, a new major version would be released). Therefore, I would expect that our [release notes for `conventional-changelog-angular`](https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-angular/CHANGELOG.md#change-log) would be sufficient to inform users of the change.

packages/conventional-changelog-angular/conventional-recommended-bump.js
hbetts: Ah, okay. I can see that. Please ignore my earlier comment.

stevemao: We can probably use https://github.com/sindresorhus/plur for pluralisation?

flxwu:
Any other issues with this PR? :smile: @hbetts @stevemao

packages/conventional-changelog-angular/conventional-recommended-bump.js
flxwu: Yeah sure, but it would be quite some work to extract all strings to be pluralized ^^ because I wouldn't just use it for this case @stevemao

packages/conventional-changelog-angular/test/test.js
```diff
@@ -72,20 +74,26 @@ describe('angular preset', function () {
       .pipe(through(function (chunk) {
         chunk = chunk.toString()
 
+        except(chunk).to.include('first build setup')
```
hbetts: This needs to be changed from `except` to `expect` (And in the other locations below)

^ Minor typo there, which is why the [CI tests are failing](https://travis-ci.org/conventional-changelog/conventional-changelog/jobs/355569196#L2236).

hbetts:
These changes look great! :+1: 

Please focus on getting the CI to pass so we can accept your work :smile:

packages/conventional-changelog-angular/test/test.js
flxwu: What about the `3) angular preset` fail? Is it also caused by this typo?

flxwu:
CI is green :+1: :smile:

hbetts:
Version [`3.0.0`](https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-angular/CHANGELOG.md#300-2018-03-22) of `conventional-changelog-angular` has been published.

packages/conventional-changelog-angular/conventional-recommended-bump.js
hbetts: I definitely think we should incorporate `plur` into `conventional-changelog` to improve messaging, but as noted by @flxwu, it's probably a significant refactor that needs to be done separately from this particular scoped change.

PR #300, opened 2018-03-28
fix: `tagPrefix` was not passed properly in conventional-changelog-core

nlunets:
Following my previous PR, I'm now looking at how to consume this properly in `standard-version` and I noticed that I forgot to pass the parameter in one place...

This PR fixes that ! The original PR can be found here https://github.com/conventional-changelog/conventional-changelog/pull/259

During the `merge-config` part of conventional-changelog-core a call is made to `gitSemverTags` to retrieve the list of tags so that the changelog can compare the previous one with the change made so far.
This call was not passing the `tagPrefix` parameter so was failing to retrieve the correct tag list !

@hbetts @bcoe please review

### DIFF
--- a/packages/conventional-changelog-core/README.md
+++ b/packages/conventional-changelog-core/README.md
@@ -115,6 +115,11 @@ Specify a package in lerna-style monorepo that the CHANGELOG should be generated
 Lerna tags releases in the format `foo-package@1.0.0` and assumes that packages
 are stored in the directory structure `./packages/foo-package`.
 
+##### tagPrefix
+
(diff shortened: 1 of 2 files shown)


hbetts:
Successfully published:
 - conventional-changelog-angular@3.0.7
 - conventional-changelog-cli@1.3.22
 - conventional-changelog-core@2.0.11
 - conventional-changelog-ember@0.3.12
 - conventional-changelog-jquery@1.3.9
 - conventional-changelog@1.1.24
 - conventional-recommended-bump@2.0.9
 - gulp-conventional-changelog@1.1.24
 - standard-changelog@1.0.19

PR #332, opened 2018-06-03
feat(preset): Angular style changelog preset for VSTS

octatone:
Initial version (fork preset and modify to work)

differences from the default Angular preset:
- issue links take the format of `host/owner/_workitems/edit/{{issue}}` (issues don't live at the repo level)
- compare links take the format of `/branches?_a=commits&baseVersion=GT{{previiousTag}}&targetVersion=GT{{currentTag}}`
- @user mention support removed (no equivalent profile links on vsts that I can find)

### DIFF
--- a/packages/conventional-changelog-angular-vsts/LICENSE.md
+++ b/packages/conventional-changelog-angular-vsts/LICENSE.md
@@ -1,0 +1,15 @@
+### ISC License
+
+Copyright © [conventional-changelog team](https://github.com/conventional-changelog)
+
+Permission to use, copy, modify, and/or distribute this software for any
+purpose with or without fee is hereby granted, provided that the above
+copyright notice and this permission notice appear in all copies.
+
+THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES WITH
+REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND
+FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT,
+INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM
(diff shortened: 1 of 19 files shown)


hbetts:
@octatone thank you for submitting this contribution.

@conventional-changelog/committee does anyone have any experience with VSTS? Is this something that we can support long-term?

Itherael:
@octatone Thanks for the contribution! My team is waiting for the new release with this feature 🥇

stevemao:
I think we should remove presets... And support `config` only. We'll maintain the popular configs.

schnitty:
I manually made some changes to do the same thing before discovering this PR.  lol
I use Azure DevOps (VSTS) extensively so will test this out and let you know how it goes.

schnitty:
@octatone What is the the preset to use this instead of angular? 
`conventional-changelog -p angular` is used for angular would it be
`conventional-changelog -p angular-vsts`  ?

stevemao:
Again I'm saying that presets was not a good idea. It bloats the package too. everything should be just pluggable config.

stevemao:
@schnitty Yes you are right, and the work hasn't started. All we need to do is to remove `preset` (docs first maybe). `--config` flag is already available though.

schnitty:
Ok, thanks @stevemao.  @Itherael if you use 
`conventional-changelog -i CHANGELOG.md -s --config conventional-changelog.js`  i
Where conventional-changelog.js is the file in the following npm module https://www.npmjs.com/package/conventional-changelog-angular-vsts
You should be golden.

stevemao:
Also, your config can just import `conventional-changelog-angular` and modify the bits you want since its very similar.

bcoe:
@octatone I'm closing this in favor of:

https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-conventionalcommits

This is a highly configurable preset, the options are outlined here: https://github.com/conventional-changelog/conventional-changelog-config-spec

bcoe:
@octatone let me know if you have any questions about how to configure this, or if we're missing any parameters you'd like to configure 👍

PR #371, opened 2018-09-08
fix(preset): ESLint recommended-bump is always "patch"

ingmarh:
Recommended version bump detection with ESLint preset is broken since
`conventional-changelog-eslint` version 3.0.0:

`npx conventional-recommended-bump -p eslint` always returns "patch".

This is because the preset's `parserOpts` (re-used with v3.0.0)
"headerCorrespondence" is set to `['tag', 'message']` rather than
`['type', 'subject']`, which was used before in the [ESLint preset
"conventional-recommended-bump" specific `parserOpts`][1].

This adapts the ESLint preset's `whatBump` implementation to check "tag"
instead of "type", which fixes the issue.

[1]: https://github.com/conventional-changelog/conventional-changelog/blob/ce1fd981f88ce201e996dfa833e4682de3aafcdd/packages/conventional-changelog-eslint/conventional-recommended-bump.js#L32-L35

### DIFF
--- a/packages/conventional-changelog-eslint/conventional-recommended-bump.js
+++ b/packages/conventional-changelog-eslint/conventional-recommended-bump.js
@@ -11,14 +11,12 @@ module.exports = {
     let features = 0
 
     commits.forEach(commit => {
-      if (!commit.type) {
-        return
-      }
+      if (!commit.tag) return
 
-      if (commit.type.toLowerCase() === 'breaking') {
+      if (commit.tag.toLowerCase() === 'breaking') {
         breakings += 1
         level = 0
-      } else if (commit.type.toLowerCase() === 'new') {
+      } else if (commit.tag.toLowerCase() === 'new') {
         features += 1
         if (level === 2) {
           level = 1


stevemao:
Thanks @ingmarh ! Feel free to add some tests...

packages/conventional-changelog-eslint/conventional-recommended-bump.js
```diff
@@ -11,14 +11,12 @@ module.exports = {
     let features = 0
 
     commits.forEach(commit => {
-      if (!commit.type) {
-        return
-      }
+      if (!commit.tag) return
```
eladchen, line 14: Under what conditions a commit has a "tag" property?

ingmarh: [`conventional-changelog-eslint` ](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-eslint) uses "tag" instead of "type" as first array entry in `headerCorrespondence` (see also the preset’s [parser-opts.js](https://github.com/conventional-changelog/conventional-changelog/blob/conventional-changelog-eslint%403.0.9/packages/conventional-changelog-eslint/parser-opts.js) and this PR’s description).

I think that’s not a mistake. The fix here only affects the ESLint preset, so you only have a "tag" property when using the ESLint preset. And not, for example, with the default preset (conventional).

eladchen: @ingmarh Got you. I was under the impression "parsed" commits properties are the same, regardless of the preset.

Issue #374, opened 2018-09-24
Tags on merge commits are ignored


Chklang:
### Version :
"conventional-changelog": "^2.0.3",

### How to call it
```
        conventionalChangelog({
            toto: 'loooooooooo',
            preset: 'angular',
            releaseCount: 0
        }).pipe...
```

### What is the problem?
When i create tags on a "merge commit", tag number is ignored. I need to remove the option "merges:false" into merge-config.js to resolve the problem. So two questions :

1. Why this option?
2. Can you add a feature to disable it if it's necessary in some cases?

Regards

dangreen:
Merge commits are configurable through commits.merges in config; the default remains false. Closing as answered.

PR #375, opened 2018-09-24
feat: Add option merges to keep merge commits #374

Chklang:
Add option "merges" into conventional options to keep merge commits into conventional-changelog process


hutson:
@conventional-changelog/conventional-changelog thoughts on this change? Seems reasonable to me.

bcoe:
@Chklang this seems reasonable to me too, sorry for dropping this on the floor for so long.

Mind rebasing with master, and we can try to land this eventually?

ahochsteger:
Any chance to get this merged?
Currently it's not possible to use conventional-changelog for releases using git-flow where releases are done using release branches that get merged into master. Since the parameter '--no-merges' is hard-coded it is not possible to find tags on merge commits.

nfantone:
@ahochsteger +1 on that. `git flow release` automatically tags merge commits and they end up being totally ignored by `conventional-changelog`.

This is the root cause of #571.

nfantone:
Would be nice to expose this as a flag in the CLI, as well.

nfantone:
I've tested this branch and, for some reason, enabling `merges` option causes the parser to ignore tags on commits that _are not merges_.

Is there any way to have _both_? Why should we even care about these details, in the first place? If a commit is tagged, just pull the tag name from it.

stevemao:
If you set `config.gitRawCommitsOpts.merge` to true, shouldn't it overwrite the default value? Maybe add a unit test to clarify?

damianopetrungaro:
It seems reasonable to me :)

nfantone:
@stevemao @damianopetrungaro Have you actually tested this? I can't make it work by setting this option alone. See #571.

zhangchao4706:
I find setting a config file which has following content makes merge commits reserved.
module.exports = {
   gitRawCommitsOpts :{
        merges: null
   }
}

wyardley:
If I hard-code to true in the current version, I get the expected set of releases, but now don't see any actual entries for the commits between tags. I think #571 may be the issue I'm seeing as well.

mediavrog:
Seems like just omitting that option would be the way to go 

https://github.com/fastlane/fastlane/blob/39afb4defa1def891506a1eaabc374ebd289e9b2/fastlane/lib/fastlane/helper/git_helper.rb#L124

mediavrog:
Started a more complete take on this issue inspired by the fastlane implementation https://github.com/conventional-changelog/conventional-changelog/pull/595

stevemao:
@nfantone you need a config file to pass option in.

stevemao:
I think instead of hacking this one, we should look at how to pass in options in general. Currently not all options are available in cli but the api is there.

mediavrog:
@stevemao Maybe this can help https://github.com/conventional-changelog/conventional-changelog/pull/595

dave-shapiro-o:
We have the same problem with tagged merge commits. Is there any new fix (E.g. by config) available in the latest version?

demirtasdurmus:
Same problem here, is there any progress on this issue?

dangreen:
You can pass `merges` option to third (`gitRawCommitsOpts`) argument of [conventionalChangelog](https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-core/index.js#L9) function and override default value

PR #377, opened 2018-09-26
fix(filter): replace `is-subset` with `lodash.ismatch`

pvdlg:
This replace [is-subset](https://github.com/studio-b12/is-subset) with [lodash.isMatch](https://lodash.com/docs/4.17.10#isMatch).

`is-subset` doesn't seems to be maintained anymore (last commit from October 2015) and it currently has a bug that prevent to use [pkg](https://github.com/zeit/pkg): https://github.com/studio-b12/is-subset/pull/9.

As a result any module that depends on `conventional-commits-filter` cannot be bundled with [pkg](https://github.com/zeit/pkg).

### DIFF
--- a/packages/conventional-commits-filter/index.js
+++ b/packages/conventional-commits-filter/index.js
@@ -1,6 +1,6 @@
 'use strict'
 
-var isSubset = require('is-subset')
+var isMatch = require('lodash.ismatch')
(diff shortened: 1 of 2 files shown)


bcoe:
@pvdlg seems reasonable 👍 thanks for the patch.

PR #380, opened 2018-10-16
fix(cli): fix issue to pass conventional-changelog standard options cli.js

jiMuBao:
fix(cli.js): fix issue where standard conventional-changelog options are not passed into the options object using --config flag


hutson:
@jiMuBao is there a reason the lock files were removed? Could they please be restored?

jiMuBao:
Hi,
Hutson, I've updaed the pull request.

Thanks

PR #385, opened 2018-10-27
Upgrade to Lerna 3, fix Node.js v11 error

evocateur:
Sorry this is huge. Probably best to avoid a squash merge, as I made my best effort to make the commits themselves atomic and self-descriptive.

## Down the Rabbit Hole

I found it very difficult to debug a [strange error](https://travis-ci.org/lerna/lerna/jobs/446305901) in the lerna integration tests when I enabled node v11 in my Travis matrix, so I upgraded the monorepo infrastructure to use relative `file:` specifiers (like Lerna itself) and run all the linting and tests (non-Windows) from the root. You know, like normal people debug.

## The _Actual_ Fix

Anyway, the _actual_ fix for my CI error is [this commit](https://github.com/conventional-changelog/conventional-changelog/commit/a6e268d22b4aa7c93b3840dd480e3443524ff112), which avoids passing falsey predicates to `Array#sort()`. Node v10 causes the tests to fail, but not the runtime (I would have hit this earlier, as I develop Lerna with Node v10). Node v11 fails _completely_, probably due to the `v8` 7.0 upgrade (significant refactoring of `Array#sort()`, ensuring stability of the sort etc).

## Minimal Acceptance

It's cool if you just want to cherry-pick a6e268d (and fefc6c8, adding node v10 to the matrix), but honestly the rest of the changes are really nice, imo. (I managed to stop myself from migrating `mocha` to `jest`, but only barely)

## Possible "Breaking" Changes

As far as I can tell, the only remotely breaking change caused by this PR would be the upgrade of the `conventional-changelog-angular` dependency in both `conventional-changelog` and `standard-changelog` from `^1.6.6` to `^5.0.1` (implicitly, during the next publish, when lerna replaces the local `file:` ref with the current version range). In my opinion, this was a bug in the current layout, as those two packages _should_ have picked up the new versions (not necessarily breaking, themselves) when `conventional-changelog-angular` was updated.

### DIFF
--- a/packages/conventional-changelog-writer/lib/util.js
+++ b/packages/conventional-changelog-writer/lib/util.js
@@ -54,14 +54,19 @@ function getCommitGroups (groupBy, commits, groupsSort, commitsSort) {
       title = false
     }
 
-    commits.sort(commitsSort)
+    if (commitsSort) {
+      commits.sort(commitsSort)
+    }
+
     commitGroups.push({
       title: title,
       commits: commits
     })
   })
 
-  commitGroups.sort(groupsSort)
+  if (groupsSort) {
+    commitGroups.sort(groupsSort)
+  }
 
   return commitGroups
 }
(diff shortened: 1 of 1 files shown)


evocateur:
Happy to help!

stevemao:
Wow, this is great! Thanks!

PR #390, opened 2018-11-06
[git-semver-tags]: BREAKING: switch params, make it normal node-style signature

tunnckoCore:
e.g. `(opts, callback)`

Also, allow passing `opts.cwd` ... because it currently fails if you want to listen on another directory

Also. I would go with `Object.assign`, but was not sure for the require node version in CI.

### DIFF
--- a/packages/git-semver-tags/index.js
+++ b/packages/git-semver-tags/index.js
@@ -1,5 +1,6 @@
 'use strict'
 
+var proc = require('process')
 var exec = require('child_process').exec
 var semverValid = require('semver').valid
 var regex = /tag:\s*(.+?)[,)]/gi
@@ -13,22 +14,19 @@ function lernaTag (tag, pkg) {
   }
 }
 
-module.exports = function (opts, callback) {
+module.exports = function gitSemverTags (opts, callback) {
   if (typeof opts === 'function') {
     callback = opts
     opts = {}
   }
-  opts = opts || {}
+  var options = Object.assign({ maxBuffer: Infinity, cwd: proc.cwd() }, opts)
 
-  if (opts.package && !opts.lernaTags) {
-    callback(Error('opts.package should only be used when running in lerna mode'))
+  if (options.package && !options.lernaTags) {
+    callback(new Error('opts.package should only be used when running in lerna mode'))
     return
   }
 
-  exec(cmd, {
-    maxBuffer: Infinity
-    cwd: opts.cwd
-  }, function (err, data) {
+  exec(cmd, options, function (err, data) {
     if (err) {
       callback(err)
       return
@@ -36,18 +34,18 @@ module.exports = function (opts, callback) {
 
     var tags = []
     var tagPrefixRegexp
-    if (opts.tagPrefix) {
-      tagPrefixRegexp = new RegExp('^' + opts.tagPrefix + '(.*)')
+    if (options.tagPrefix) {
+      tagPrefixRegexp = new RegExp('^' + options.tagPrefix + '(.*)')
     }
     data.split('\n').forEach(function (decorations) {
       var match
       while ((match = regex.exec(decorations))) {
         var tag = match[1]
-        if (opts.lernaTags) {
-          if (lernaTag(tag, opts.package)) {
+        if (options.lernaTags) {
+          if (lernaTag(tag, options.package)) {
             tags.push(tag)
           }
-        } else if (opts.tagPrefix) {
+        } else if (options.tagPrefix) {
           var matches = tag.match(tagPrefixRegexp)
           if (matches && semverValid(matches[1])) {
             tags.push(tag)
--- a/packages/git-semver-tags/index.js
+++ b/packages/git-semver-tags/index.js
@@ -13,7 +13,11 @@ function lernaTag (tag, pkg) {
   }
 }
 
-module.exports = function (callback, opts) {
+module.exports = function (opts, callback) {
+  if (typeof opts === 'function') {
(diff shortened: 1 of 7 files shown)


tunnckoCore:
And still. Even with `cwd` for some reason `git log` does not found that it is a git repo.

tunnckoCore:
Updated

```js
'use strict';

import proc from 'process';
import { exec } from '@tunnckocore/execa';
import semver from 'semver';

const regex = /tag:\s*(.+?)[,)]/gi;
const cmd = 'git log --decorate --no-color';

function lernaTag(tag, pkg) {
  if (pkg && !new RegExp(`^${pkg}@`).test(tag)) {
    return false;
  }

  // eslint-disable-next-line unicorn/no-unsafe-regex
  return /^.+@\d+\.\d+\.\d+(-.+)?$/.test(tag);
}

/* eslint-disable no-param-reassign */
export default async function gitSemverTags(options) {
  const opts = Object.assign({ cwd: proc.cwd() }, options);

  if (opts.package && !opts.lernaTags) {
    throw new Error(
      'opts.package should only be used when running in lerna mode',
    );
  }

  const [{ stdout: data }] = await exec(cmd, opts);

  const tags = [];
  let tagPrefixRegexp;
  if (opts.tagPrefix) {
    tagPrefixRegexp = new RegExp(`^${opts.tagPrefix}(.*)`);
  }
  data.split('\n').forEach((decorations) => {
    let match;
    /* eslint-disable no-cond-assign */
    while ((match = regex.exec(decorations))) {
      const tag = match[1];
      if (opts.lernaTags) {
        if (lernaTag(tag, opts.package)) {
          tags.push(tag);
        }
      } else if (opts.tagPrefix) {
        const matches = tag.match(tagPrefixRegexp);
        if (matches && semver.valid(matches[1])) {
          tags.push(tag);
        }
      } else if (semver.valid(tag)) {
        tags.push(tag);
      }
    }
  });

  return tags;
}

```

let me now if you want it that way directly.

tunnckoCore:
I'll fix the PR, there is missing `,` too.

packages/git-semver-tags/index.js
```diff
@@ -13,7 +13,11 @@ function lernaTag (tag, pkg) {
   }
 }
 
-module.exports = function (callback, opts) {
+module.exports = function (opts, callback) {
```
evocateur: I think you would need to modify all the callers of this method among the sibling packages, as well? Probably the tests, too.

tunnckoCore: Oooh. Yea, sorry.

tunnckoCore: Actually, tests are good. Will add one more for when there is `opts`.

tunnckoCore:
Also. I reviewed the other packages and only two of them are using it, neither of them needs an update. They are using `gitSemverTags(cb)`, which is supported here too.

But anyway, I'll add test and update to use `Object.assign`.

tunnckoCore:
Are you okey switching to Sindre's `execa`?

tunnckoCore:
I don't think Promise is needed. Everyone can promisify a callback nowadays.
Also, it seems that the other packages are using callbacks too so not make much sense to turn that into a Promise API.

bcoe:
@tunnckoCore mind rebasing? happen to be doing some work on this repo and noticed lots of stale pull requests (sorry!).

tunnckoCore:
I'll try in one two days. (setuping the dev machine)

bcoe:
@tunnckoCore is this still something you would like to land?

tunnckoCore:
Uuuh, sorry. Of course.

I'll be up and running in next week.

packages/git-semver-tags/index.js
```diff
@@ -13,36 +14,38 @@ function lernaTag (tag, pkg) {
   }
 }
 
-module.exports = function (callback, opts) {
-  opts = opts || {}
+module.exports = function gitSemverTags (opts, callback) {
```
hutson: :+1: for correct callback order

packages/git-semver-tags/index.js
```diff
@@ -13,36 +14,38 @@ function lernaTag (tag, pkg) {
...
+  if (typeof opts === 'function') {
+    callback = opts
+    opts = {}
+  }
+  var options = Object.assign({ maxBuffer: Infinity, cwd: proc.cwd() }, opts)
```
hutson:
```javascript
const options = { maxBuffer: Infinity, cwd: proc.cwd(), ...opts}
```

We can use object spread now that the minimum Node version we support is Node 8 LTS, assuming we don't need to rely on the behavior of `Object.assign` to call setters (also, I changed `var` to `const`)

tunnckoCore: cool cool. I should read about the diffs again, any suggestions? I started using it everywhere and still prefer it over spread for some unknown reason. ;d

tunnckoCore:
Let me know if any other thing is needed :)

packages/git-semver-tags/index.js
hutson: > I should read about the diffs again, any suggestions?

I tend to just read [Mozilla documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax). :sweat_smile:  In my experience their examples are pretty good.

> I started using it everywhere and still prefer it over spread for some unknown reason. ;d

That's totally fine. My suggestion, after further consideration, was purely based on  aesthetics (from my perspective).

PR #394, opened 2018-11-09
fix(preset:angular): Scoped NPM packages are not seen as GitHub username anymore

Kocal:
GitHub username detection of Angular preset detects scoped NPM packages as GitHub username, package `@dummy/package` will be replaced by `[@dummy](https://github.com/dummy)/package` in the changelog.

selection_999 155

With this patch, scoped NPM packages won't be detected as GitHub username:
selection_999 157

### DIFF
--- a/packages/conventional-changelog-angular/writer-opts.js
+++ b/packages/conventional-changelog-angular/writer-opts.js
@@ -79,7 +79,13 @@ function getWriterOpts () {
         }
         if (context.host) {
           // User URLs.
-          commit.subject = commit.subject.replace(/\B@([a-z0-9](?:-?[a-z0-9]){0,38})/g, `[@$1](${context.host}/$1)`)
+          commit.subject = commit.subject.replace(/\B@([a-z0-9](?:-?[a-z0-9/]){0,38})/g, (_, username) => {
+            if (username.includes('/')) {
+              return `@${username}`
+            }
+
+            return `[@${username}](${context.host}/${username})`
+          })
         }
       }
 
--- a/packages/conventional-changelog-angular/test/test.js
+++ b/packages/conventional-changelog-angular/test/test.js
@@ -56,6 +56,7 @@ betterThanBefore.setups([
   },
   function () {
     gitDummyCommit(['fix: use npm@5 (@username)'])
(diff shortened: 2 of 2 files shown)


evocateur:
This has impacted me too, thanks for taking the time to figure it out!

Kocal:
Me too I spent time trying modify the regex by adding `(?!/)` at the end, but it didn't work very well.. 😒

stevemao:
Sorry, just wanted another eye on this

packages/conventional-changelog-angular/writer-opts.js
```diff
@@ -79,7 +79,13 @@ function getWriterOpts () {
         }
         if (context.host) {
           // User URLs.
-          commit.subject = commit.subject.replace(/\B@([a-z0-9](?:-?[a-z0-9]){0,38})/g, `[@$1](${context.host}/$1)`)
+          commit.subject = commit.subject.replace(/\B@([a-z0-9](?:-?[a-z0-9/]){0,38})/g, (_, username) => {
+            if (username.includes('/')) {
+              return `@${username}`
```
stevemao: I'm wondering if it'd be good if we also link the project. But it won't be generic enough (wouldn't work if its not a scoped module).

Kocal:
No problem, thanks for the merge :)

packages/conventional-changelog-angular/writer-opts.js
Kocal: That's really to hard to make it work for a not scoped module.

There is a _solution_, but it's not bullet-proof. 
On some of my projects, I use [Dependabot](https://dependabot.com/) which updates my dependencies automatically by doing PR like this:

selection_340

Here, if we can detect `bump PACKAGE_NAME from X to Y`, then we can link the project even if it's not a scoped module. 
A more generic solution is to detect only `PACKAGE_NAME from X to Y`.

An other solution would be to detect all valid packages name in the commit subject, and then check for `https://registry.npmjs.org/:packageName`. But that means that those packages will be matched too: 
- https://registry.npmjs.org/bump
- https://registry.npmjs.org/from
- https://registry.npmjs.org/to

So I don't think we can make this... :disappointed:

stevemao: Yeah, not a good idea :)

PR #395, opened 2018-11-30
feat(parser): parse merge message and conventional commit on same line

cgosiak:
Can now parse both a merge message and a conventional commit that appear on the same line

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 **/node_modules/
 **/tmp/
 /lerna-debug.log
+.idea
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "test-windows": "lerna exec --concurrency 1 -- npm run test-windows",
     "release": "lerna publish --conventional-commits"
   },
-  "version": "1.0.0",
+  "version": "1.0.1",
   "devDependencies": {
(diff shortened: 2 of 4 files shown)


stevemao:
Hi @cgosiak , thanks for the PR. If you want to generate a changelog based on pull-request messages, you could have a look at how [ember](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-ember) does it. Let me know if that solves your problem.

cgosiak:
So ember seems to do something closer to what I am looking for. However, the changes that I propose would make it so the "Merge Pattern" and a Conventional Commit may reside on the same line. I know that this is the case for what I am dealing with now; but it seems like it could be a generic enough enhancement that other variations may find value in it.

cgosiak:
@stevemao any thoughts?

bcoe:
@cgosiak I'm not a fan of this being in the base library; if you want to handle the edge case of commits rewritten with a `Merge ` in the commit line, perhaps create your own preset?

bcoe:
@cgosiak please feel free to start a conversation here:

http://devtoolscommunity.herokuapp.com/

I think this could potentially be a candidate for writing your own preset.

PR #409, opened 2019-02-13
feat: send options to whatBump

avaly:
In custom scenarios it is useful to get some context of the current options of the `conventionalRecommendedBump` call.
    
For example, when using a custom changelog preset together with Lerna, and the preset uses a custom `whatBump` method that requires to know the current Lerna package that it is being run for.
    
See: https://github.com/lerna/lerna/blob/a6733a2b864cf9d082d080bbd3bfedb04e59b0ab/core/conventional-commits/lib/recommend-version.js#L13-L21

### DIFF
--- a/packages/conventional-recommended-bump/index.js
+++ b/packages/conventional-recommended-bump/index.js
@@ -73,7 +73,7 @@ function conventionalRecommendedBump (optionsArgument, parserOptsArgument, cbArg
             warn(`No commits since last release`)
           }
 
-          let result = whatBump(commits)
+          let result = whatBump(commits, options)
 
           if (result && result.level != null) {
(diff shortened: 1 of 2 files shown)


hutson:
@avaly your change looks fine with me, but may I request you add your pull request comment, along with a concrete example of how `options` would be used, to your commit message. (It would also help to incorporate that concrete example to the unit tests to ensure our tests are checking that the concrete example works with this change)

I want to make sure that the reasoning for this change is well documented. It will help us know why we're supporting it and why we need to keep supporting it in the future.

cc @conventional-changelog/conventional-changelog for an additional pair of eyes on this change.

avaly:
@hutson Thanks! I've updated the PR description and the commit message to include the specific use case this was meant for.

Issue #317, opened 2018-04-10
Question: How to configure "angular" preset to print docs, style, etc.


MaiKaY:
Hi guys,
i made a lot of research and even sniffed the repositories to find a proper solution but it doesn't work out for me. Therefore i decided to ask here for help.

My goal is to `generateNotes` even for commits prefixed with `docs:`, `style:`, `ci:`, etc.
Right now only `fix:`, `feat:` and `breacking change` are printed to `CHANGELOG.md` and GitHub releases.

My configuration looks like this:

```
{
    "release": {
        "verifyConditions": [
            "@semantic-release/changelog",
            "@semantic-release/github"
        ],
        "analyzeCommits": {
            "preset": "angular",
            "releaseRules": [
                {
                    "type": "docs",
                    "release": "patch"
                },
                {
                    "type": "refactor",
                    "release": "patch"
                },
                {
                    "type": "style",
                    "release": "patch"
                }
            ]
        },
        "generateNotes": {
            "preset": "angular"
        },
        "prepare": [
            "@semantic-release/changelog",
            "@semantic-release/git"
        ],
        "publish": [
            "@semantic-release/github"
        ]
    }
}
```
I know that `angular` is the default preset for `generateNotes` but i added it to be explicit for everyone.

I appreciate every single hint, link or solution snippet.

Thanks in advance.

MaiKaY:
Hi @hbetts,
sorry to poke you... but could you support me in this case?

damianopetrungaro:
Hey, sorry for the delay @MaiKaY 
IMHO all the types should be printed in the changelog, our [specification](https://conventionalcommits.org/) is pretty free about it.

Are you able to prepare a PR?
Maybe @bcoe  know better than me 😄

MaiKaY:
Hi @damianopetrungaro,
thanks a lot for your answer!

I my point of view the issue is at this specific point: https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-angular/writer-opts.js#L44-L45

I created now a `release.config.js` with the following `writerOpts.transform`:
```
'transform': (commit, context) => {
    if (commit.type === 'feat') {
        commit.type = 'Features';
    } else if (commit.type === 'fix') {
        commit.type = 'Bug Fixes';
    } else if (commit.type === 'perf') {
        commit.type = 'Performance Improvements';
    } else if (commit.type === 'revert') {
        commit.type = 'Reverts';
[...]

PR #410, opened 2019-02-13
fix(angular): do not skip commit types

Visya:
Before most commit types were skipped and not written into the changelog. This change allows all commit types be part of the changelog.

This was discussed in the issue https://github.com/conventional-changelog/conventional-changelog/issues/317, but no changes were made.

### DIFF
--- a/packages/conventional-changelog-angular/writer-opts.js
+++ b/packages/conventional-changelog-angular/writer-opts.js
@@ -25,12 +25,10 @@ module.exports = Q.all([
 function getWriterOpts () {
   return {
     transform: (commit, context) => {
-      let discard = true
       const issues = []
 
       commit.notes.forEach(note => {
         note.title = `BREAKING CHANGES`
-        discard = false
       })
 
       if (commit.type === `feat`) {
@@ -41,8 +39,6 @@ function getWriterOpts () {
         commit.type = `Performance Improvements`
       } else if (commit.type === `revert`) {
         commit.type = `Reverts`
-      } else if (discard) {
-        return
       } else if (commit.type === `docs`) {
         commit.type = `Documentation`
       } else if (commit.type === `style`) {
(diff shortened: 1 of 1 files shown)


hutson:
@Visya looking through [Angular's changelog](https://github.com/angular/angular.js/blob/master/CHANGELOG.md) I don't see CI, Docs, or other commit types published. Does this change break how Angular currently generates their changelog?

cc @damianopetrungaro I believe this need is completely legitimate, but maybe not for the Angular configuration. What are your thoughts on how we can meet this need?

Visya:
Hi @hutson 

You're right, Angular doesn't include anything except `feat`, `fix` and `revert` in their changelog unless it's a breaking change (https://github.com/angular/angular/issues/5672).

My reasoning was that it's nice to see all changes associated with the release, but on the other hand it can be seen as redundant for the user to see changes in the changelog that do not impact functionality. Maybe, a way to go would be to add this as a configuration option?

JeffBeltran:
wanted to chime in that i would love a "config" option that allows us to override the defaults

stevemao:
I don't think we should change angular config. We can create a new config.

myii:
@Visya Thanks for this.  We've been using this fix in [saltstack-formulas/template-formula/release.config.js](https://github.com/saltstack-formulas/template-formula/blob/c71915ed1e38d72c38360d409ddcffd01a61f557/release.config.js#L23-L104) and it is working perfectly.

hutson:
I feel the answer right now is to leave the `angular` preset unmodified, and encourage anyone wanting a slightly different output to create a fork of the `angular` preset with their changes.

Long-term, @bcoe is looking at supporting preset configuration in #421. A change that will hopefully meet your needs @Visya

bcoe:
@hutson I agree, I'm hoping that the `conventional-changelog-conventionalcommits` preset will create a good foundation for an easier to config `preset` that adheres explicitly to conventionalcommits.org.

stevemao:
Closing in favour of https://github.com/conventional-changelog/conventional-changelog/pull/421

PR #412, opened 2019-02-14
chore: update handlebars dependency

bcoe:
addresses https://www.npmjs.com/advisories/755

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "pretest": "npm run lint",
     "test": "nyc mocha --timeout 30000 packages/*/test{,/*}.js",
     "test-windows": "lerna exec --concurrency 1 -- npm run test-windows",
-    "release": "lerna publish --conventional-commits"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -30,7 +30,7 @@
     "pretest": "npm run lint",
     "test": "nyc mocha --timeout 30000 packages/*/test{,/*}.js",
     "test-windows": "lerna exec --concurrency 1 -- npm run test-windows",
-    "release": "lerna publish --conventional-commits"
+    "release": "lerna publish --conventional-commits --npm-tag=next"
```
bcoe: switch to `next` tag, to make rollout easier (OTP sometimes bites us).

evocateur: Alternatively, could use the [`--temp-tag`](https://github.com/lerna/lerna/tree/master/commands/publish#--temp-tag) flag for extra insurance (without leaving the extra dist-tag hanging around in the event of success).

Issue #413, opened 2019-02-21
fix(git-raw-commits): conflicting "path" and custom args 


Akuukis:
`git-raw-commits` puts all custom git args after path, which are not picked up due `--` seperator added together with path. 

## Observed
```
git log --format=%B%n-hash-%n%H%n------------------------ >8 ------------------------ v1.0.0..HEAD -- workspaces/myPackage/ --first-parent
```
## Expected
```
git log --format=%B%n-hash-%n%H%n------------------------ >8 ------------------------ v1.0.0..HEAD --first-parent -- workspaces/myPackage/
```

stevemao:
I'm not following... Would you be able to send a PR with failing tests?

Akuukis:
@stevemao Makes sense, here's the failing test in PR.

dangreen:
The Git client supports firstParent and places it before the path separator; arbitrary raw arguments are no longer a public API. Closing as resolved.

PR #416, opened 2019-02-27
test(git-raw-commits): git parameter order

Akuukis:
Failing test for #413

### DIFF
--- a/packages/git-raw-commits/test.js
+++ b/packages/git-raw-commits/test.js
@@ -192,4 +192,21 @@ describe('git-raw-commits', function () {
         done()
       }))
   })
+
+  it('should add all git parameters before "-- <path>" part for them not to be ignored', function (done) {
+    var command = ''
+
+    gitRawCommits({
+      firstParent: true,
+      path: './packages/foo',
+      debug: (output) => command = output.split('\n').pop(),
+    })
+      .pipe(through(function (chunk, enc, cb) {
+        cb()
+      }, function () {
+        expect(command).to.match(/git log --format=%B%n------------------------ >8 ------------------------ --first-parent HEAD -- .\/packages\/foo/)
+        done()
+      }))
+  })
+
 })


bcoe:
@Akuukis I don't 100% follow what bug this is introducing for you?

Akuukis:
@bcoe if I pass `firstParent` to params, it doesn't get applied. E.g.
```ts
     gitRawCommits({
      firstParent: true,
      path: './packages/foo',
      debug: (output) => command = output.split('\n').pop(),
    })
```

It's because it ends up in wrong location on the command.

FYI My workaround for now:
```ts

const gitRawCommitsCustom = (from: string = '', path?: string) => {
    // TODO: Use upstream after fix: https://github.com/conventional-changelog/conventional-changelog/issues/413
    // Original params:
    // {
    //     firstParent: true,
    //     format: `%B%n-hash-%n%H`,
    //     from,
    //     ...(path ? {path} : {}),
    //     debug: console.log.bind(console),
    // }

    // tslint:disable: no-unsafe-any only-arrow-functions no-empty ban
    const execFile = require('child_process').execFile
    const split = require('split2')
    const stream = require('stream')
    const through = require('through2')

    const DELIMITER = '------------------------ >8 ------------------------'
    const readable = new stream.Readable()
    readable._read = function() {}

    let isError = false

    const child = execFile('git', [
            `log`,
            `--format=%B%n-hash-%n%H%n${DELIMITER}`,
            `${from}..HEAD`,
            `--first-parent`,
            ...(path ? ['--', path] : []),
        ], {
            maxBuffer: Infinity,
        })

    child.stdout
        .pipe(split(`${DELIMITER}\n`))
        .pipe(through(function(chunk, enc, cb) {
            readable.push(chunk)
            isError = false

            cb()
        }, function(cb) {
            setImmediate(function() {
                if (!isError) {
                readable.push(null)
                readable.emit('close')
                }

                cb()
            })
        }))

    child.stderr
        .pipe(through.obj(function(chunk) {
        isError = true
        readable.emit('error', new Error(chunk))
        readable.emit('close')
        }))
    // tslint:enable: no-unsafe-any only-arrow-functions no-empty

    return readable
}
```

bcoe:
@Akuukis what git workflow do you use that this feature helps address?

I'm open to this feature, if you would like to put the work in to implement it :+1: Just want to make sure we've documented appropriately why the parameter exists.

Akuukis:
@bcoe I'm not requesting a new feature, but reporting a bug. The config value already existed there, but silently fails to do anything. So we either fix the config value to do what it's supposed to do or remove it.

The [first result on search](https://marcgg.com/blog/2015/08/04/git-first-parent-log/) explains well how `firstParent` is useful. Furthermore, it simplifies logic when operation on `master` or `dev` branches - it's easier to handle few big merge commits than many small commits.

PR #420, opened 2019-03-10
docs: add vscode extension link to README.md

axetroy:
It's a part of conventional-changelog ecosystem


stevemao:
Thank you @axetroy , could you add it https://github.com/conventional-changelog/conventional-changelog/blob/master/README.md instead please?

PR #422, opened 2019-03-18
feat: .changelogrc

jbottigliero:
- Adds two example fixtures for `.changelogrc`

---

This is a WIP PR. I mostly wanted to make sure I was on the same page as the update suggested in #421. 

Work can either be continued here or closed out.

### DIFF
--- a/packages/conventional-changelog-core/test/fixtures/changelogrc/.changelogrc
+++ b/packages/conventional-changelog-core/test/fixtures/changelogrc/.changelogrc
@@ -1,0 +1,3 @@
+{
+    "extend": ["preset:conventionalcommits"]
+}
--- a/packages/conventional-changelog-core/test/fixtures/changelogrc/.changelogrc-extend
+++ b/packages/conventional-changelog-core/test/fixtures/changelogrc/.changelogrc-extend
@@ -1,0 +1,6 @@
+{
+    "writerOpts": {
+        "host": "https://hosted-scm.biz"
+    },
+    "extend": ["preset:conventionalcommits", "../invalid/local/path/.changelogrc"]
+}


jbottigliero:
**`preset:` prefix in `extend`**
This seemed like the best option to both specify a current preset and/or extend a preset.

Alternative 1: `name`
```js
{
  "name": "conventionalcommits"
}
```
This didn't seem declarative enough and could be potentially confusing when used in conjuntion with `extend`.

Alternative 2: `preset.name: <string>` or `preset: <string>`
```js
{
  "preset": "conventionalcommits"
}
// or 
{
  "preset": {
     "name":  "conventionalcommits"
  }
}
```
A _bit_ more declarative in the `preset: <string>` form, but still might cause confusion with `extend`. The object form seems like a non-start, as we basically want the top-level object to represent a "preset".

Alternative 3: `extend` Without Identifier
```js
{
  "extend": ["conventionalcommits"]
}
```
Without the identifier in `extend` it would require additional parsing to determine if it's an attempt to load a local file or a named preset. Additionally, we'd have to make a decision extend precedence, which could be avoided...

bcoe:
rather than adding to the fixtures directory, I would be tempted to fork https://github.com/Jimdo/changelogrc-spec into a new directory in the `packages/` folder (with `publish: false` set in the package.json); alternatively, I could create a new top level repo for you to commit to.

I think that the `changelogrc-spec`, from my point of view, has the right idea of the level of granularity folks will want to modify -- I started implementing support for the configuration here:

https://github.com/conventional-changelog/conventional-changelog/pull/421/files#diff-3ac2e1c1a67d64c40f587c73b9ed4811R125

Introducing a few more configuration options that allow users to edit partials inside templates:

`issueUrlFormat`, `compareUrlFormat`, my thinking being this would be an easy way for someone to support gitlab format rather than, as an example, GitHub format.

----

I think that whether or not this configuration is in a `.rc` file is an implementation detail, which I'd leave to the upstream implementer like `standard-version` ... maybe we should call the spec:

*conventional-changelog-config-spec*?

bcoe:
@jbottigliero I think how this would change your work in https://github.com/conventional-changelog/standard-version/pull/278, would simply be that you pass a different config object to the new `conventional-changelog-conventionalcommits` preset.

bcoe:
@jbottigliero I've created a repo here, for us here:

https://github.com/conventional-changelog/conventional-changelog-config-spec

to rough out what we think we should expose for configuration, let's move the conversation there.

jbottigliero:
Catching up on all the latest updates and will head over to the new repository... thanks for triaging!

PR #434, opened 2019-04-10
feat: conventionalcommits preset, preMajor config option

bcoe:
## What is this?

This PR introduces the `preMajor` config option for the conventionalcommits preset.

When set to `true`, breaking changes will still be populated in `reasons` but a `minor` version bump will be recommended rather than a `major`.

----

when landing #421 I neglected to pass the new preset object to recommended bump; this PR:

1. Pulls the logic into `conventional-changelog-preset-loader` so that it's shared by both `conventional-changelog` and recommended bump.
2. introduces the `preMajor` config option ([documented here](https://github.com/conventional-changelog/conventional-changelog-config-spec/pull/4)); which is required by `standard-version`, and is used in unit-tests to ensure the preset is getting passed configuration.


bcoe:
going ahead and landing this, since it's partially a fix for #421 and it would be nice to get this configuration work finally published to npm.

PR #437, opened 2019-04-23
fix: Downgrade node 10.x dependency to 6.9.0 dependency

### LINKED ISSUE #436: conventional-changelog-conventionalcommits requires Node 10?
I recently started seeing [CI failures](https://travis-ci.org/omghax/ember-mock-socket-shim/jobs/519188552#L518) in a project that uses [`standard-version`](https://github.com/conventional-changelog/standard-version), seemingly because `conventional-changelog-conventionalcommits` [requires Node 10 or greater](https://github.com/conventional-changelog/conventional-changelog/blob/e84a647b5884eaa3eb6324853fd026c29629e600/packages/conventional-changelog-conventionalcommits/package.json#L28). The output from Yarn is:

```
error conventional-changelog-conventionalcommits@1.1.1: The engine "node" is incompatible with this module. Expected version ">=10.0.0". Got "6.17.1"
```

The same thing happens locally if I try to run `yarn upgrade standard-version`. The other packages in this repo specify `>= 6.9.0` in their `engines` field. Is there a reason this one requires a different version?

tmcw:
Fixes #436.

Tested with 6.9.0, and it works.

### DIFF
--- a/packages/conventional-changelog-conventionalcommits/package.json
+++ b/packages/conventional-changelog-conventionalcommits/package.json
@@ -25,7 +25,7 @@
   ],
   "author": "Ben Coe",
   "engines": {
-    "node": ">=10.0.0"
+    "node": ">=6.9.0"
   },
   "license": "ISC",
   "bugs": {


bcoe:
@tmcw thanks for sending the patch :+1: I'll release a patch today.

PR #440, opened 2019-04-24
fix: template should not include multiple H1 tags

bcoe:
CHANGELOGS should not contain multiple H1 tags.


sumigoma:
@bcoe I presume this was done to allow users to specify their own `header:` config option, but `writer-opts.js` doesn't seem to be using `config.header` and adding the config option doesn't seem to actually do anything.

This change makes `conventional-changelog-conventionalcommits` inconsistent with all the other presets and I would prefer to only conditionally convert <h1> to <h2> tags if a custom header is provided. Would you be ok if I created a PR to do this (and also add support for config.header which appears to be missing)?

PR #443, opened 2019-05-02
feat: ! without BREAKING CHANGE should be treated as major

bcoe:
BREAKING CHANGE: if `!` is in the commit header, it now indicates a BREAKING CHANGE, and the description is used as the body.


cgatian:
@bcoe may I ask where this change comes from? I dont see this in [the specification](https://www.conventionalcommits.org/en/v1.0.0-beta.2/#specification)?

cgatian:
@bcoe Sorry I see now that this is a part of the 1.0.0 spec: https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-both--and-breaking-change-footer

PR #452, opened 2019-05-06
fix: Recommend a patch bump for features when preMajor is enabled

coreyfarrell:


evocateur:
What was the context for this change?

coreyfarrell:
@evocateur conventional-changelog/standard-version#308 is the original ticket I opened.  The idea is if someone has a dependency on `"@myorg/module": "^0.1.0"`, this should allow all non-breaking changes to be installed (thus a non-breaking feature/fix release after 0.1.0 would be 0.1.1).  This is especially important when dealing with bug fixes, you want to avoid bumping to 0.2.0 for non-breaking changes as this will force all dependencies to update in order to get the bug fix.

evocateur:
@coreyfarrell Ah, thanks! That makes total sense.

PR #463, opened 2019-05-15
fix(preset, conventionalcommits): Ensure proper substitutions for the conventionalcommit preset by using commit context for values where possible.

### LINKED ISSUE #462: (preset) conventionalcommits : Cross-Repository References
The `conventional-changelog-conventionalcommits` preset seems to be ignoring cross-repository references. I _believe_ this is caused by the Handlebar templates being used for the new spec substitutions not account for context.

I could see this also being the cause of conventional-changelog/standard-version#358.

jbottigliero:
- adds conditionals around commit context for host, owner and repository.

closes #462

### DIFF
--- a/packages/conventional-changelog-conventionalcommits/test/test.js
+++ b/packages/conventional-changelog-conventionalcommits/test/test.js
@@ -1,6 +1,7 @@
 'use strict'
 var conventionalChangelogCore = require('conventional-changelog-core')
-var preset = require('../')()
+var getPreset = require('../')
+var preset = getPreset()
 var expect = require('chai').expect
 var mocha = require('mocha')
 var describe = mocha.describe
@@ -156,7 +157,7 @@ describe('conventionalcommits.org preset', function () {
       }))
   })
 
-  it('should properly format cross-repository issues', function (done) {
+  it('should properly format external repository issues', function (done) {
     preparing(1)
     conventionalChangelogCore({
       config: preset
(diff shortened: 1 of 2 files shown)


bcoe:
@jbottigliero this looks good :+1: is there any way to add a slightly more thorough test, or perhaps could we add one upstream in `standard-version`?

jbottigliero:
@bcoe – I'm not super familiar with Handlebars... but I was mostly basing this off of the [`context` being passed along](https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-core/lib/merge-config.js#L168) and modified throughout the `conventional-changelog(-writer)` code and [eventually passed to the compiled template](https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-writer/lib/util.js#L192).

It doesn't look like any of the other handlebar templates worry about a contextualized host so maybe it's not necessary? ...I should mention when I was working on this fix I assumed conventional-changelog/standard-version#358 was reproducible.

bcoe:
@jbottigliero if we can't reproduce, I'd be tempted to perhaps abandon this PR; since this introduces a bit more complexity than the other templates.

Any objection?

jbottigliero:
I _think_ there is still some version of the PR that needs to exist... basically, I'm trying to ensure [`issueUrlFormat`](https://github.com/conventional-changelog/conventional-changelog/blob/fix-conventionalcommits-substitutions/packages/conventional-changelog-conventionalcommits/templates/commit.hbs#L22) contains similar logic to the [`angular`](https://github.com/conventional-changelog/conventional-changelog/blob/fix-conventionalcommits-substitutions/packages/conventional-changelog-angular/templates/commit.hbs#L34-L53) implementation.

At the moment, the cross-repository test _will_ fail by defaulting to `@root.owner` and `@root.repository`.

I don't think any of the current templates use `this.host`, so that one I can certainly drop.

bcoe:
@jbottigliero I'll trust what you decide on this PR, but let's just make sure there's a reproduction for any of the edge-cases we think we're solving for.

Just let me know when you feel this PR is ready?

jbottigliero:
@bcoe this should be good to go and it will resolve #462.

The `conventionalcommits` preset still does not support `repoUrl`, but looking a little closer it's only really used by the [`angular` preset](https://github.com/conventional-changelog-archived-repos/conventional-changelog-writer/commit/e504682), most other presets override the `commit.hbs` _without_ referencing the `repoUrl`... and seems like something we should just move away from, relying strictly on the configuration spec for conf values or any sort of fallbacks required.

bcoe:
@jbottigliero thank you :+1:

PR #468, opened 2019-05-22
fix(deps): update conventional-changelog-core dependency get-pkg-repo…

ryancat:
… to v4 (#467)


nmccready:
Why was this toss aside with zero comments?

PR #478, opened 2019-06-09
fix(Delimiter): Breaks debug output.

### LINKED ISSUE #477: [git-raw-commits] Delimiter breaks debug output.
See this line:

https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/git-raw-commits/index.js#L10

When debug is enabled, a console output of what arguments is being sent to `git log` is printed out, including the hard coded delimter `------------------------ >8 ------------------------`

For example:

```
Your git-log command is:
git log --format=%B%n-hash-%n%H%n-gitTags-%n%d%n-committerDate-%n%ci%n-authorName-%n%an%n-authorEmail-%n%ae%n-gpgStatus-%n%G?%n-gpgSigner-%n%GS%n------------------------ >8 ------------------------ v3.17..v3.18 --no-merges
```

If you actually copy that command and try run it yourself, you get this error: ❌
``` 
fatal: unrecognized argument: ------------------------
```

If you edit it, and remove the to spaces either side of the `>8`, and run the command, no output at all is generated. ❌

If you remove the delimiter, the command works, and git log outputs the log as formatted, between the two tags requested. ✅ 

If you change the delimiter to ANYTHING not including hyphens, ie `_______>8_______` then the command also works. ✅

I'm not sure why the delimiter would be fine when run during node and then break during terminal execution, but it makes debug a hassle as I always have to copy the command and remove the delimiter.

I'm running git version 2.21.0 on zsh macos.

samkelleher:
Fixes #477 

Fixes issue as described

What the delimiter looks like with change:

```
git log --format=%B%n-hash-%n%H%n-gitTags-%n%d%n-committerDate-%n%ci%n-authorName-%n%an%n-authorEmail-%n%ae%n-gpgStatus-%n%G?%n-gpgSigner-%n%GS%n____________________________________________________ git-semver-tags@2.0.0..git-semver-tags@2.0.3
```

What it looked like before the change:
```
git log --format=%B%n-hash-%n%H%n-gitTags-%n%d%n-committerDate-%n%ci%n-authorName-%n%an%n-authorEmail-%n%ae%n-gpgStatus-%n%G?%n-gpgSigner-%n%GS%n------------------------ >8 ------------------------ git-semver-tags@2.0.0..git-semver-tags@2.0.3
```

The original version fails with `fatal: unrecognized argument: ------------------------`, the fixed version executes normally.

Run this `git log` command above in the `conventional-changelog` repo itself to see the change.

### DIFF
--- a/packages/git-raw-commits/index.js
+++ b/packages/git-raw-commits/index.js
@@ -7,7 +7,7 @@ var stream = require('stream')
 var template = require('lodash.template')
 var through = require('through2')
 
-var DELIMITER = '------------------------ >8 ------------------------'
+var DELIMITER = '____________________________________________________'
 
 function normalizeExecOpts (execOpts) {
   execOpts = execOpts || {}


bcoe:
@samkelleher to get the git command working, it just needs to have quotes around the format, like so:

```
git log --format="%B%n-hash-%n%H%n-gitTags-%n%d%n-committerDate-%n%ci%n-authorName-%n%an%n-authorEmail-%n%ae%n-gpgStatus-%n%G?%n-gpgSigner-%n%GS%n------------------------ >8 ------------------------ v3.17..v3.18" --no-merges
```

I'd prefer if we address this issue by adding quotes, rather than changing the delimiter; as changing the delimiter would be a breaking change across all dependent libraries.

Would happily accept a patch with the modified debug output however 😄

PR #480, opened 2019-06-09
feat(gitExecOpts): Provide facility to define options.

### LINKED ISSUE #479: [changelog-core] Cannot pass `rawExecOpts` to git-raw-commits.
See line:

https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-core/index.js#L25

`git-raw-commits` accepts two config params, `rawGitOpt` and `rawExecOpts`, however, the latter is not passed so thus a consumer of changelog-core cannot pass these options.

In my use case, I am running the changelog against a different working directly rather than assuming that the target source code being examined by the changelog tool is run within the current same directly.

samkelleher:
Fixes #479 

This allows a new param to be provided to core that are passed to git-raw-commit as exec option, notable, now being able to change the cwd from a calling function.

If no value provided, it'll work normally as `cwd` is defaulted to the `process.cwd()` value when falsy.

### DIFF
--- a/packages/conventional-changelog-core/index.js
+++ b/packages/conventional-changelog-core/index.js
@@ -6,7 +6,7 @@ const conventionalChangelogWriter = require('conventional-changelog-writer')
 const stream = require('stream')
 const through = require('through2')
 const mergeConfig = require('./lib/merge-config')
-function conventionalChangelog (options, context, gitRawCommitsOpts, parserOpts, writerOpts) {
+function conventionalChangelog (options, context, gitRawCommitsOpts, parserOpts, writerOpts, gitRawExecOpts) {
   writerOpts = writerOpts || {}
 
   var readable = new stream.Readable({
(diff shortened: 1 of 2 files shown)


samkelleher:
`git-semver-tags` package already has this change in 2.0.3, however, only 2.0.2 is published. Can someone publish 2.0.3 - then the `cwd` can be changed. In 2.0.2, the cwd is not configurable. 2.0.3 introduced the change in #390

bcoe:
@samkelleher `git-semver-tags@2.0.3` is now latest. Would it be possible for you to add a couple unit tests for this change? that way we can confirm things are working as expected.

bcoe:
@samkelleher :wave: just wanted to follow up and see if this is still something you'd like to land. You should be able to add some unit tests now.

vivaxy:
It would be better if `gitRawExecOpts` is passed from `conventional-changelog` module to `conventional-changelog-core`.

PR #481, opened 2019-06-12
Allow use of full package names when loading presets

eemeli:
It's rather surprising and confusing esp. with scoped packages that the CLI effectively requires using an option like `--preset=@scope/preset` in order to fetch the preset defined in `@scope/conventional-changelog-preset`.

This change allows for the loader to also handle full package paths, so that with the above example `--preset=@scope/conventional-changelog-preset` would work.


bcoe:
this is a great improvement; agreed this was very confusing 👍

PR #518, opened 2019-09-03
fix(preset-loader): fix handling conventionalcommits preset without config object

tommywo:
fix(preset-loader): fix handling conventionalcommits preset without config object

### DIFF
--- a/packages/conventional-changelog-preset-loader/index.js
+++ b/packages/conventional-changelog-preset-loader/index.js
@@ -33,8 +33,9 @@ function presetLoader (requireMethod) {
       // rather than returning a promise, presets can return a builder function
       // which accepts a config object (allowing for customization) and returns
       // a promise.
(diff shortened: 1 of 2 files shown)


packages/conventional-changelog-preset-loader/index.js
```diff
@@ -35,6 +35,10 @@ function presetLoader (requireMethod) {
       // a promise.
       if (config && !config.then && typeof path === 'object') {
         return config(path)
```
jbottigliero: Rather than the additional condition, should we just alter the condition that determines whether or not it's a builder?

```js
if (config && !config.then) {
  const options = typeof path === 'object' ? path : {}
  return config(options)
}
```

tommywo: Yes, that looks much better. Thanks.

PR #525, opened 2019-09-26
fix: revertPattern in angular preset should also match regular revert…

YuraBeznos:
… format

Part of conventional-changelog/issues/348

### DIFF
--- a/packages/conventional-changelog-angular/parser-opts.js
+++ b/packages/conventional-changelog-angular/parser-opts.js
@@ -8,6 +8,6 @@ module.exports = {
     `subject`
   ],
   noteKeywords: [`BREAKING CHANGE`],
-  revertPattern: /^revert:\s([\s\S]*?)\s*This reverts commit (\w*)\./,
+  revertPattern: /^(?:Revert|revert:)\s"?([\s\S]+?)"?\s*This reverts commit (\w*)\./,
   revertCorrespondence: [`header`, `hash`]
 }


tommywo:
Hi @YuraBeznos. Thank you for contributing. 🎉  
Can you add a unit test for that?

stevemao:
Should conventional-changelog-conventionalcommits use the default revert message and leave the angular preset as is?

tommywo:
@stevemao https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#revert now doesn't say the message format is autogenerated, but it uses word SHOULD not MUST, so I personally think default message format should be also parsed in angular preset.

tommywo:
@bcoe, @jbottigliero, @byCedric  what's your opinion on this? this might affect how commitlint is working after https://github.com/conventional-changelog/commitlint/pull/821 is merged

bcoe:
should we just make the regex case-insensitive? thoughts @tommywo?

byCedric:
@tommywo I don't think this is an issue in Commitlint, it's backwards compatible 😄 (at least, I think it is?) We also keep track of some "ignored commits", like reverts ([which is case-insensitive already](https://github.com/conventional-changelog/commitlint/blob/master/%40commitlint/is-ignored/src/defaults.ts#L22))

tommywo:
closing in favor of #545

PR #564, opened 2019-12-02
test: make all tests run on windows

tommywo:


stevemao:
Great work! Thanks for the effort!

.travis.yml
```diff
@@ -15,14 +18,15 @@ cache:
     - ${HOME}/.npm
 
 before_install:
-  - npm install --global npm@latest
```
stevemao: I'm wondering if it's necessary to do this on Mac or Linux? If not, we can remove the whole thing.

package.json
```diff
@@ -95,5 +98,16 @@
...
     "standard-changelog": "file:packages/standard-changelog"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "*.js": [
+      "eslint --fix --cache --cache-location ./node_modules/.cache/ --ignore-path .gitignore",
```
stevemao: TIL: `--ignore-path .gitignore` nice trick :)

.travis.yml
tommywo: I think that was included to use `npm ci`, so if we test only node version that have npm >=5.7.0 it should be safe to remove.

packages/conventional-changelog-jquery/test/test.js
```diff
@@ -19,19 +19,12 @@ describe('jquery preset', function () {
...
-    shell.exec('git add --all && git commit -m"Event: Remove an internal argument to the on method" -m"Fixes #2, #4, gh-200"')
-    writeFileSync('test4', '')
-    shell.exec('git add --all && git commit -m"Manipulation: Remove an internal argument to the remove method" -m"Closes #22"')
-    writeFileSync('test5', '')
-    shell.exec('git add --all && git commit -m"Bad commit"')
-    writeFileSync('test6', '')
-    shell.exec('git add --all && git commit -m"Core: Create jQuery.ajax" -m"Closes gh-100"')
+    gitDummyCommit(['Core: Make jQuery objects iterable'])
+    gitDummyCommit(['CSS: Don\'t name the anonymous swap function'])
+    gitDummyCommit(['Event: Remove an internal argument to the on method" -m"Fixes #2, #4, gh-200'])
```
tommywo:
```suggestion
    gitDummyCommit(['Event: Remove an internal argument to the on method', 'Fixes #2, #4, gh-200'])
```

packages/conventional-changelog-jquery/test/test.js
```diff
@@ -19,19 +19,12 @@ describe('jquery preset', function () {
...
+    gitDummyCommit(['Manipulation: Remove an internal argument to the remove method" -m"Closes #22'])
```
tommywo:
```suggestion
    gitDummyCommit(['Manipulation: Remove an internal argument to the remove method', 'Closes #22'])
```

packages/conventional-changelog-jquery/test/test.js
```diff
@@ -19,19 +19,12 @@ describe('jquery preset', function () {
...
+    gitDummyCommit(['Bad commit'])
+    gitDummyCommit(['Core: Create jQuery.ajax" -m"Closes gh-100'])
```
tommywo:
```suggestion
    gitDummyCommit(['Core: Create jQuery.ajax', 'Closes gh-100'])
```

.travis.yml
stevemao: Let's remove it then :)

PR #569, opened 2019-12-10
feat: support slash in headerPattern default options

noelebrun:
Currently, if you're using a "/" character in the scope of your commit and using the default option for "headerPattern", the Regex doesn't match the scope and the commit is considered as invalid.

This pull request adds the "/" character to the default "headerPattern" option.

### DIFF
--- a/packages/conventional-commits-parser/index.js
+++ b/packages/conventional-commits-parser/index.js
@@ -7,7 +7,7 @@ var _ = require('lodash')
 
 function assignOpts (options) {
   options = _.extend({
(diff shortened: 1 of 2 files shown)


bcoe:
@no23reason 👌 thanks for the contribution. It might be smart for us to better define the range of characters for each component on conventionalcommits.org

no23reason:
@bcoe The credit is not mine but @noelebrun's , you probably made a typo :)

bcoe:
@noelebrun, what @no23reason said 😜

PR #570, opened 2019-12-10
fix(conventional-changelog-core): fix duplicated commits when `from` is specified

### LINKED ISSUE #567: Duplicated entries for commits
Example here: https://travis-ci.org/conventional-changelog/standard-version/jobs/622392913#L342

This bug was introduced in 2fba5c7a02e0e34093a6bd9a01109457db9b84c5 
#408

Affected packages: 

> standard-changelog@2.0.19,2.0.20
> gulp-conventional-changelog@2.0.27,2.0.28
> conventional-changelog@3.1.16,3.1.17
> conventional-changelog-core@4.1.2,4.1.3
> conventional-changelog-cli@2.0.29,2.0.30

cc: @HitkoDev

HitkoDev:
fixes #567

### DIFF
--- a/packages/conventional-changelog-core/index.js
+++ b/packages/conventional-changelog-core/index.js
@@ -51,7 +51,7 @@ function conventionalChangelog (options, context, gitRawCommitsOpts, parserOpts,
         const from = i > 0
           ? reverseTags[i - 1]
           : gitRawCommitsOpts.from || ''
-        if (gitRawCommitsOpts.from) {
+        if (gitRawCommitsOpts.from && gitRawCommitsOpts.from !== from) {
           let hasData = false
           return commitsRange(gitRawCommitsOpts.from, to)
             .on('data', function () {
--- a/packages/conventional-changelog-core/test/test.js
+++ b/packages/conventional-changelog-core/test/test.js
@@ -803,6 +803,28 @@ describe('conventionalChangelogCore', function () {
       }))
   })
 
+  it('should read each commit range exactly once', function (done) {
+    preparing(9)
+
+    conventionalChangelogCore({
+      preset: {
+        compareUrlFormat: '/compare/{{previousTag}}...{{currentTag}}'
+      }
+    }, {}, {}, {}, {
+      headerPartial: '',
+      commitPartial: '* {{header}}\n'
+    })
+      .pipe(through(function (chunk, enc, cb) {
+        chunk = chunk.toString()
+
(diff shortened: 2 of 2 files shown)


tommywo:
this is still not right when release count is set to more then 1. 
so with history right that:
Screenshot 2019-12-11 at 21 04 18
when I run
`cli.js -p angular -i CHANGELOG.md -s -r 2`
it will run git log with ranges:
```
v1.0.2..v1.0.0
v1.0.2..v1.0.1
v1.0.2..v1.0.2
v1.0.2..v1.0.3
v1.0.2..HEAD
v1.0.3..HEAD
```
instead of
```
v1.0.2..v1.0.3
v1.0.3..HEAD
```

tommywo:
Something like that will work
```javascript
      var reverseTags = context.gitSemverTags.slice(0).reverse()
      reverseTags.push('HEAD')

      if (gitRawCommitsOpts.from) {
        // we are only generating part of changelog so we ignore part of tags
        reverseTags = reverseTags.slice(reverseTags.indexOf(gitRawCommitsOpts.from))
        // TODO: handle reverseTags.indexOf(gitRawCommitsOpts.from) == -1
        // Can gitRawCommitsOpts.from be set to something that is not semver tag but different ref? 
        // if yes we should find first tag after that ref, 
        // slice reverseTags from there and add that strange at the begging of  reverseTags
        // if not we should log an error end return
       }

      var commitsErrorThrown = false

      var commitsStream = new stream.Readable({
        objectMode: true
      })
      commitsStream._read = function () { }

      function commitsRange (from, to) {
        return gitRawCommits(_.merge({}, gitRawCommitsOpts, {
          from: from,
          to: to
        }))
          .on('error', function (err) {
            if (!commitsErrorThrown) {
              setImmediate(commitsStream.emit.bind(commitsStream), 'error', err)
              commitsErrorThrown = true
            }
          })
      }
      
      var streams = reverseTags.map((to, i) => {
        const from = i > 0
          ? reverseTags[i - 1]
          : ''
        return commitsRange(from, to)
      })

      if (gitRawCommitsOpts.from) {
        // we are only generating part of a changelog so we skip
        // git log   `..gitRawCommitsOpts.from`
        streams = streams.splice(1)
      }

      if (gitRawCommitsOpts.reverse) {
        streams.reverse()
      }
```

tommywo:
closing in favor of #573

PR #580, opened 2019-12-26
feat: add issuePrefixesCaseSensitive parser option

kjr-stratslab:
Currently issuePrefixes are always treated as case insensitive. 

This is problematic for JIRA-type prefixes (i.e. {some-letters}-) as it creates a lot of false positives. Example: `upgrade dependency to foobar-12` causes foobar-12 to be treated as a reference.

This MR adds an option to control that behaviour. The default is `false` so it is backwards compatible.


bcoe:
@kjr-stratslab thanks for the contribution 👌

GuoXiaoyang:
So when will you publish a new version which contains this pr?
I noticed that the latest version of `conventional-commits-parser` is 3.0.8 which is published 5 months ago.

Issue #13, opened 2014-06-10
Option for commits patterns


ocombe:
I would like to be able to use a custom commit pattern. It would be nice to have an option to customize the following regex:
```js
var COMMIT_PATTERN = /^(\w*)(\(([\w\$\.\-\* ]*)\))?\: (.*)$/;
```

And also I would like to be able to list more than just `feat` & `fix` commits in the changelog (`docs` for example, and probably other things later), it would be nice to be able to provide a list of the headers to parse.

If you don't have the time to do it but think it's a good idea, just tell me and I'll work on it and send a PR.

stevemao:
Also a custom closes pattern would be nice

https://help.github.com/articles/closing-issues-via-commit-messages/

ocombe:
Woot!! \o/
Thanks a lot Steve!

stevemao:
Please checkout the rewrite on the master branch and see if the problem still exists. Thanks.

PR #586, opened 2020-01-15
feat: allows notes pattern to be customized

dnlfrst:
# What is the problem?

Inspired by #13, I would like to customize the [`RegEx` pattern](https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-commits-parser/lib/regex.js#L21) used for notes similar to the option that is [already available for the header pattern](https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-commits-parser/lib/parser.js#L169). Using the note keywords does not suffice.

# What could be a solution?

Add an option `notesPattern` that takes a function `(noteKeywords: string) => RegExp`. With this, a user can specify a custom regular expression that incorporates the (already concatenated) note keywords.

### DIFF
--- a/packages/conventional-commits-parser/README.md
+++ b/packages/conventional-commits-parser/README.md
@@ -205,6 +205,12 @@ Type: `array` of `string` or `string` Default: `['BREAKING CHANGE']`
 
 Keywords for important notes. This value is case **insensitive**. If it's a `string` it will be converted to an `array` separated by a comma.
 
+##### notesPattern
+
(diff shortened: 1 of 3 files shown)


tommywo:
Hi @dnlfrst and thank you for making a pull request. 🎉 
Could you also add that options to the documentation?

dnlfrst:
> Could you also add that options to the documentation?

I am happy to do so but only as early as at the end of next week.

dnlfrst:
@tommywo I have added the option to the documentation, this PR is ready to go!

bcoe (approved):

bcoe (approved):

PR #593, opened 2020-02-12
Support loose semver option for SemVer package

ruhrohraggy:
### The Problem
If using versions that do not match the standard semantic version format, tags do not get recognized correctly so that changes may be collected to be written into the changelog. For example, for a version 1.0.01-1, SemVer's `valid` function fails the version, and therefore the tag isn't seen as a tag to reference for releases.

### The Solution
The SemVer package allows for loose validation for verisons that are close to a semantic version, but do not quite match their stricter pattern. With these changes, the CLI can be invoked with an option passed that will allow for versions to be validated with a 'loose' regex.


packages/conventional-changelog-writer/README.md
```diff
@@ -85,7 +85,7 @@ Type: `string`
 
 Type: `boolean` Default: `semver.patch(context.version) !== 0`
 
-By default, this value is true if `version`'s patch is `0`.
+By default, this value is true if `version`'s patch is not `0`.
```
ruhrohraggy: Unless I was reading this incorrectly, seemed like this was a typo. `isPatch` should be false if the patch is 0, not the other way around

ruhrohraggy:
Forgive me if I've broken any PR etiquette. I probably should have created a branch on this repo rather than try to merge master of my fork. I look forward to review!

packages/conventional-changelog-writer/index.js
```diff
@@ -30,7 +30,11 @@ function conventionalChangelogWriter (context, options) {
     noteGroupsSort: 'title',
     notesSort: 'text',
     generateOn: function (commit) {
-      return semverValid(commit.version)
+      var semverOptions = {}
```
ruhrohraggy: There is another option available for SemVer functions, but loose was the only applicable one. In future work, we could support more than just `loose`

ruhrohraggy:
Apologies for the multiple commits -- this is now fully functionally tested and ready for review (and hopefully merge and release)

ruhrohraggy:
I'm done now I promise :) Lots of additions here. Let me know if you'd prefer they be broken up -- they're all just changes I need soonish and I'd love to not have to fork and maintain my own version of this awesome package. Thanks guys

ruhrohraggy:
@tommywo @stevemao (Only tagging you guys because it seems you've reviewed some stuff most recently) Let me know if I need to do anything to break this PR down into smaller changes. I know there's a lot more lines changed here than in a lot of the PR's that are opened (at least one's that I've seen). Thanks guys!

ruhrohraggy:
Hey folks, just checking if anyone's had a chance to look at this. I hate to be a nag.

packages/conventional-changelog-writer/index.js
```diff
@@ -30,7 +30,11 @@ function conventionalChangelogWriter (context, options) {
...
+      if (options.looseSemver) {
+        semverOptions = { loose: true }
```
hutson: Does this check that `options.looseSemver` is not `null`?

Could you do this in one line with `semverOptions = { loose: options.looseSemver ? true : false }`

packages/conventional-changelog-writer/test/index.spec.js
```diff
@@ -305,6 +305,7 @@ describe('conventionalChangelogWriter', function () {
         }))
         .pipe(through(function (chunk, enc, cb) {
           expect(chunk.toString()).to.contain('# 1.0.0 ')
+          console.log(chunk.toString())
```
hutson: Was this just for testing? Can it be removed?

packages/conventional-changelog-writer/test/index.spec.js
```diff
@@ -851,6 +852,70 @@ describe('conventionalChangelogWriter', function () {
...
+        notes: [],
+        references: [],
+        committerDate: '2015-04-07 15:01:30 +1000'
+      })
+      upstream.end()
+
+      return upstream
+    }
+
+    it('should generate on `\'version\'` if it\'s at least a valid loose semver and loose option is passed', function (done) {
```
hutson: Thank you for including unit tests!! :+1:

packages/conventional-changelog-writer/test/index.spec.js
```diff
@@ -851,6 +852,70 @@ describe('conventionalChangelogWriter', function () {
...
+      var i = 0
+
+      getStream()
+        .pipe(conventionalChangelogWriter({ version: '1.1.01-1' }, { looseSemver: true }))
```
hutson: I haven't read through this code in a while, so I may be missing this: Is there a test that validates expected behavior when `version: '1.1.01-1'` is set and `looseSemver` is `false`?

packages/conventional-changelog-writer/index.js
ruhrohraggy: It doesn't - it'll just fall out and `semverOptions` will never be set which is null-safe in the semver package where the options are passed. Happy to clean it up though

packages/conventional-changelog-writer/test/index.spec.js
ruhrohraggy: Yes good catch, I'll remove

packages/conventional-changelog-writer/test/index.spec.js
ruhrohraggy: Absolutely! I'm not going to try to get away with adding this much logic without testing it 🤣

packages/conventional-changelog-writer/test/index.spec.js
ruhrohraggy: Let me dig into it again and figure out. I think passing the version directly to the writer will still allow the full version to be written (with the leading zero in this case), it'll just fail validation for seeing if it's a patch version.

ruhrohraggy:
Hey guys, just checking in to see if you could take another look over this. Thanks!

stevemao:
I like this option to be added to the writer package. But I'm not sure if we should add it to the cli since cli can take `writeOpts`

ruhrohraggy:
> I like this option to be added to the writer package. But I'm not sure if we should add it to the cli since cli can take `writeOpts`

That would require extending an existing preset though, wouldn't it? Making it available via the CLI just allows for more flexibility depending on who's using it. My personal use case, we have a defined preset with writerOpts that comes from another third party, and it'd be ideal if we didn't have to extend it or customize it to pass the option in, but instead just pass the option from where the whole process is invoked, the CLI.

stevemao:
I understand, but hard to pass all options through cli. Need to find a better way to do this.

ruhrohraggy:
If it's alright with you @stevemao, I'd like to hear others' thoughts on it. I personally feel like adding an extra option to the CLI is simply extending the interface and exposing the **full** functionality of the package, which to me is the point of the CLI. More importantly (and the main reason I was making these changes), the option is also used for changing how semantic tags are collected. I don't think it would make sense for the `git-semver-tags` package to read in an option from `writerOpts`. The use of the option in the writer is really just for detecting the nature of the version for how it's written.

@hutson @tommywo what are your thoughts?

packages/conventional-changelog-cli/cli.js
```diff
@@ -46,8 +46,14 @@ var cli = meow(`
                                 Example of a config script: https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-cli/test/fixtures/config.js
 
       -c, --context             A filepath of a json that is used to define template variables
+
       -l, --lerna-package       Generate a changelog for a specific lerna package (:pkg-name@1.0.0)
+
       -t, --tag-prefix          Tag prefix to consider when reading the tags
+
+      -l, --loose-semver        Validate with semver with a loose option - allowing flexibility in the version format
```
tommywo: shortflag `-l` is already taken by lerna-package

ruhrohraggy: I'm not sure how I missed that - I'll remove the shortflag option and just leave `--loose-semver`

ruhrohraggy:
Just checking back in on this - we'd love to be able to get it in an official release and make use of it. Thanks guys!

ruhrohraggy:
Hey guys - just checking in again. I'm sure things are pretty chaotic in all of our worlds. Feel free to reach out if you need anything from me to move this along

ruhrohraggy:
Hey guys, is there any way we can get this moving along? Anything you guys need from me? It'd be super helpful for my team not to have to fork and maintain this whole repo. Thanks! And stay safe!

ruhrohraggy:
Hey friends. Just checking on this one more time. Thanks for any attention you can give to it. @hutson @tommywo @stevemao

ruhrohraggy:
@bcoe @stevemao @hutson @tommywo Any update on this guys? It's kind of been forgotten.

bcoe:
@ruhrohraggy could I bother you to rebase, sorry that this got dropped on the floor.

bcoe:
@ruhrohraggy bother you to rebase, if this is still something you'd like to land.

dangreen:
Closing this stale PR because the requested rebase was not provided and the implementation no longer applies cleanly to the current package layout. If support for non-standard versions is still needed, please open a fresh, focused issue or PR against the current codebase with a minimal reproduction.

PR #595, opened 2020-02-14
feat(options): add mergeCommitFilter option to cli

mediavrog:
Based off the discussion on https://github.com/conventional-changelog/conventional-changelog/pull/375 and other issues, I started an implementation to add a new option to either include, exclude or only show merge commits when generating the changelog. 

It still needs tests and I need to test the code myself properly, but so far I could already verify that the changelog is complete when just dropping the `merges:false` from the `gitRawCommitOpts` in `merge-config.js`.

For now, I think the option `merge-commit-filter` should default to `exclude`, as this is the current behavior and would not break existing usage.

I will try to clean things up later when I have more time, but please also feel free to contribute.

### DIFF
--- a/packages/conventional-changelog-core/lib/merge-config.js
+++ b/packages/conventional-changelog-core/lib/merge-config.js
@@ -69,7 +69,6 @@ function mergeConfig (options, context, gitRawCommitsOpts, parserOpts, writerOpt
       }
     },
     append: false,
-    mergeCommitFilter: 'exclude',
     releaseCount: 1,
     debug: function () {},
     transform: function (commit, cb) {
@@ -241,6 +240,7 @@ function mergeConfig (options, context, gitRawCommitsOpts, parserOpts, writerOpt
       gitRawCommitsOpts = _.assign({
         format: '%B%n-hash-%n%H%n-gitTags-%n%d%n-committerDate-%n%ci',
         from: fromTag,
+        merges: false,
         debug: options.debug
       },
       config.gitRawCommitsOpts,
@@ -249,16 +249,6 @@ function mergeConfig (options, context, gitRawCommitsOpts, parserOpts, writerOpt
 
       if (options.append) {
         gitRawCommitsOpts.reverse = gitRawCommitsOpts.reverse || true
-      }
-
-      if (options.mergeCommitFilter) {
-        if (options.mergeCommitFilter === 'exclude') {
-          gitRawCommitsOpts.merges = false
-        } else if (options.mergeCommitFilter === 'only-merges') {
-          gitRawCommitsOpts.merges = true
-        } else if (options.mergeCommitFilter === 'include') {
-          // noop; just not set `merges` option in rawCommitOpts
-        }
       }
 
       parserOpts = _.assign(
--- a/packages/conventional-changelog-core/lib/merge-config.js
+++ b/packages/conventional-changelog-core/lib/merge-config.js
@@ -69,6 +69,7 @@ function mergeConfig (options, context, gitRawCommitsOpts, parserOpts, writerOpt
       }
     },
     append: false,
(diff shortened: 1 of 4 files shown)


packages/conventional-changelog-core/lib/merge-config.js
```diff
@@ -251,6 +251,16 @@ function mergeConfig (options, context, gitRawCommitsOpts, parserOpts, writerOpt
         gitRawCommitsOpts.reverse = gitRawCommitsOpts.reverse || true
       }
 
+      if (options.mergeCommitFilter) {
+        if (options.mergeCommitFilter === 'exclude') {
+          gitRawCommitsOpts.merges = false
```
tommywo: If one would set `gitRawCommitsOpts.merges` to `null` or `true`  it would get overwritten to `false`.

tommywo: I think it should be moved to `conventional-changelog-cli`, as people using `conventional-changelog-core` directly can set `gitRawCommitsOpts.merges` and this is only needed for command line flags.

packages/conventional-changelog-core/lib/merge-config.js
```diff
@@ -251,6 +251,16 @@ function mergeConfig (options, context, gitRawCommitsOpts, parserOpts, writerOpt
...
+        } else if (options.mergeCommitFilter === 'only-merges') {
+          gitRawCommitsOpts.merges = true
+        } else if (options.mergeCommitFilter === 'include') {
+          // noop; just not set `merges` option in rawCommitOpts
```
tommywo: If this is a no-op then there is no need for `if`.

tommywo (changes requested):
Hi @mediavrog and thank you for contributing.  🎉 
I've added some notes on the code.

packages/conventional-changelog-core/lib/merge-config.js
mediavrog: got it. I will remove it from the core then

packages/conventional-changelog-core/lib/merge-config.js
mediavrog: Have a look at https://github.com/conventional-changelog/conventional-changelog/commit/ac1f189550e1acc92f77e1b4fdcbd8581c10b317

mediavrog:
verified working using the below command on a project using git-flow (and thus tags on merge commits)

```
npm i -g ./packages/standard-changelog
standard-changelog --first-release -m include
standard-changelog --first-release -m only-merges
standard-changelog --first-release -m exclude
standard-changelog --first-release -m abc // default to exclude
```

mediavrog:
@tommywo I need some help with the tests. If you could have a look at https://github.com/conventional-changelog/conventional-changelog/pull/595/files#diff-177d48f386b5a5dce18e8dd697f40b1dR304 and point me to what I am doing wrong.

Also, I'd ideally want to make the tests for `----merge-commit-filter` independent from `--first-release`

Kubessandra (approved):

Kubessandra:
Any updates about this PR? Need this for the repo I'm using.

bcoe:
@mediavrog I apologize for the slow momentum on this project recently. Could I bother you to resolve the conflicts on this branch?

RangerRick:
I tried doing a quick merge to see if I could get this working but the tests fail for me and I don't understand the internals enough to fix it. If someone has more time than I to look into it, my attempts at cleanup are [here](https://github.com/RangerRick/conventional-changelog/tree/include-merge-commits)

JulianSanio:
Can anybody help out and support?

dangreen:
Thanks for the implementation. The current generator supports merge selection through the commits.merges configuration accepted by --config, and this PR targets the retired CLI layout. Closing this older implementation as superseded.

Issue #596, opened 2020-02-16
conventional-recommended-bump does not calculate lerna packages


jjangga0214:
`conventionalRecommendedBump` does not work with `lernaPackage`. 

For example, the result would be same for '@foo/bar1',  '@foo/bar2', and  '@foo/bar3', no matter what commits they are related.

Of course, `lerna version`'s result is different, which conventionalRecommendedBump should be same with.

I may add an repro repo if you request.

```ts
import conventionalRecommendedBump from 'conventional-recommended-bump'

for (const pkgName of [
      '@foo/bar1',
      '@foo/bar2',
      '@foo/bar3',
    ]) {
      conventionalRecommendedBump(
        {
          preset: `angular`,
          lernaPackage: pkgName,
        },
        (error, recommendation) => {
          if (error) {
            throw new Error(error)
          }
          console.log(`${pkgName}: ${recommendation.releaseType}`) 
        },
      )
    }
```

tommywo:
Hi @jjangga0214 
Lerna uses `conventional-recommended-bump` under the hood. https://github.com/lerna/lerna/blob/master/core/conventional-commits/lib/recommend-version.js#L47 You are missing `path` options that would be passed to gitRawCommitsOpts https://github.com/conventional-changelog/conventional-changelog/blob/49dcd6e1e37e5834e48ca56bdf7b74b718251f38/packages/conventional-recommended-bump/index.js#L75

jjangga0214:
@tommywo 

It works like a charm :)
 
Thank you so much!

PR #597, opened 2020-02-17
docs(conventional-recommended-bump/README): add options.path

jjangga0214:
As you can see #596, I think `options.path` is good to be documented.

Thanks!


tommywo:
Thank you very much @jjangga0214 🥇 🎉

PR #602, opened 2020-02-20
fix(preset:angular): escaped scoped string should not be seen as GitHub username (#601)

### LINKED ISSUE #601: Angular preset: avoid faulty username replace
Hello everyone! 😊 

My issue is linked to #218 but it's not exactly the same.
With the current implementation https://github.com/conventional-changelog/conventional-changelog/blob/b9f50573f292ea29ff51627646ca7825bf182d52/packages/conventional-changelog-angular/writer-opts.js#L82

A fix comment like ``fix(ui): remove incorrect `@internal` from comments of property`` will be parsed as ``fix(ui): remove incorrect `[@internal](https://github.com/internal)` from comments of property``. 

I think that we should follow, as much as possible, the markdown implementation of GitHub and not interprating strings between `` ``.

In this case, the message should be parsed as: ``fix(ui): remove incorrect `@internal` from comments of property`` since the `@internal` is escaped.

What do you think ? 😊

SuperITMan:
ISSUES CLOSED: #601

### DIFF
--- a/packages/conventional-changelog-angular/test/test.js
+++ b/packages/conventional-changelog-angular/test/test.js
@@ -57,6 +57,9 @@ betterThanBefore.setups([
   function () {
     gitDummyCommit(['fix: use npm@5 (@username)'])
     gitDummyCommit(['build(deps): bump @dummy/package from 7.1.2 to 8.0.0', 'BREAKING CHANGE: The Change is huge.'])
+    gitDummyCommit(['fix: remove \\`@internal\\` in comments of property'])
+    gitDummyCommit(['fix(test): fix \\`code example\\` by @dlmr'])
+    gitDummyCommit(['feat: adding support for \\` by @superitman!'])
   },
   function () {
     gitDummyCommit(['Revert \\"feat: default revert format\\"', 'This reverts commit 1234.'])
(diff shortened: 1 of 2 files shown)


SuperITMan:
@dudekaa @bcoe could you please have a look at this? 😊

dudekaa:
Hi @SuperITMan 

thanks for contributing...but as first note: I'm no more attached to this project than you, so for valid answers you have to ask project maintainers. I can make only advices here!

Now, I feel you' re patching this at wrong place at all, so it gets easily over-comlicated... as I understand your bug report, you simply want to teach this RegExp
```javascript
commit.subject = commit.subject.replace(/\B@([a-z0-9](?:-?[a-z0-9/]){0,38})/g, (_, username) => {
```
to simply don't match quoted user refs and you should be fine... It' s much more simpler and understandable approach...

But once more again, I' m just contributor as you, so you can't take me for granit.

BR
Arnost

bcoe:
@SuperITMan sorry that we dropped this PR on the floor. @dudekaa I agree that I think we could figure out a slight rework of `commit.subject.replace` and address the problem.

SuperITMan:
Hello @bcoe @dudekaa 
Thanks for the remark, I'll have a look in the next days to improve this :-)

SuperITMan:
I updated a bit my PR and added some comments to explain why the check is there.
With simple regex, there is unfortunately no way to check the number of occurrences of a character (in this case: `).

If you have a better solution in mind, please, feel free to share it :blush:

SuperITMan:
@bcoe I updated my PR. Could you have a look at this ?

bcoe:
@SuperITMan thanks for updating the PR 👍 will try to help merge some of these PRs soon.

dangreen:
Thanks for the original fix. The issue has since been resolved by #1444: mention linkification now skips inline-code spans, with regression coverage. Since #601 is closed as completed, this older PR is superseded. Closing as completed.

PR #598, opened 2020-02-17
various fixes regarding `gitRawExecOpts`

kellyselden:

PR #604, opened 2020-02-25
fix: pass gitRawExecOpts to gitRawCommits

eunjae-lee:
Hi all,

There's a missing argument.


eunjae-lee:
oh, it's a duplicate. #598

PR #611, opened 2020-03-11
chore(conventional-changelog-writer): bump handlebars to 4.5.3

frassinier:
`handlebars` is vulnerable to arbitrary code execution. The vulnerability exists as the lookup helper does not properly validate templates, allowing the execution of JavaScript code in templates.

Just bump it to `^4.5.3`


bcoe:
addressing this in a larger update of deps to address a few issue. Thanks for the patch 👍

PR #618, opened 2020-04-05
fix: pass config to parserOpts and writerOpts

bertho-zero:


bcoe:
@bertho-zero what's broken without this, we only get the default configuration I imagine?

Could I bother you to add a test?

MunifTanjim:
> what's broken without this, we only get the default configuration I imagine?

Yes, without this, it's not possible to use custom configuration. I'm using this change in my project (https://github.com/Synor/synor/blob/71ddc337/patches/conventional-changelog-conventionalcommits%2B4.2.3.dev.patch) and can confirm that it works perfectly.

acostalima:
Hi guys,
When would it be possible to get this merged? We're having issues with [lerna](https://github.com/lerna/lerna/issues/2138#issuecomment-584468055) when using `conventionalcommits` preset.

CC @satazor

PR #624, opened 2020-04-16
Bugfixes for conventional-changelog-{conventionalcommits,preset-loader}

MunifTanjim:
Fixes some subtle bugs.

### DIFF
--- a/packages/conventional-changelog-conventionalcommits/index.js
+++ b/packages/conventional-changelog-conventionalcommits/index.js
@@ -8,22 +8,22 @@ const writerOpts = require('./writer-opts')
 
 module.exports = function (parameter) {
   // parameter passed can be either a config object or a callback function
+  let config = parameter
+  let callback
+
   if (_.isFunction(parameter)) {
-    // parameter is a callback object
-    const config = {}
-    // FIXME: use presetOpts(config) for callback
-    Q.all([
-      conventionalChangelog(config),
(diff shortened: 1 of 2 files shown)


bertho-zero:
https://github.com/conventional-changelog/conventional-changelog/pull/618

MunifTanjim:
Oh, sorry 😅 that overlapped.

bcoe:
@MunifTanjim I went ahead and landed #618, is there still anything from this PR that we need?

MunifTanjim:
> Is there still anything from this PR that we need?

- A refactor for the loader `conventional-changelog-conventionalcommits` (nice to have)
- A bugfix for `conventional-changelog-preset-loader`

bcoe:
@MunifTanjim 👍 any tests we could add for the `FIXME`, that you can think of?

spooks me a little bit just because it removes a lot of code (but perhaps we're covered by existing test cases).

MunifTanjim:
The removed codes were basically duplicate code. And while doing this refactor I indeed broke & fixed the existing tests.

> any tests we could add for the FIXME, that you can think of?

But there are no current test for the scenario when custom config is passed. And custom config wasn't even working before #618 . So, maybe we should add something for that?

But I think that's a separate task.

MunifTanjim:
Rebased on latest `master`

PR #669, opened 2020-08-27
feat(conventionalcommits): allow matching scope

ext:
Allow matching scope when generating changelog. For instance `chore(deps)` could be included in changelog under its own section while other `chore` is hidden.


OlofFredriksson (approved):

bcoe (approved):

bcoe (approved):
This seems like a useful feature, thanks for the contribution 👍

bcoe:
@ext thank you for the contribution 👌

felipecrs:
There is an issue in the config-spec repo for this feature: https://github.com/conventional-changelog/conventional-changelog-config-spec/issues/53

I guess it's only missing the new schema now.

Issue #213, opened 2017-07-22
Parse failing with git config showSignature


Conaclos:
Hi!

# Symptom

I noted the generation of empty changelog. e.g. [this](https://github.com/Conaclos/eslint-config-conaclos/blob/master/CHANGELOG.md).

# Issue

The problem is coming of my git configuration. Indeed, if I remove the display of commit signatures, then the changelog is properly generated

```
[log]
    showSignature = true
```

thesmiley1:
I think I have managed to run into this same issue using `standard-version` (maybe this bug report belongs there?).  What _seems_ (I haven't traced code) to be happening is that `standard-version` (silently) ignores commits that it fails to parse (as (presumably) happens in this case, choking on the signature).  So effectively all signed commits are ignored when `log.showSignature` is enabled.

It can be a good thing to ignore parsing errors from repositories that didn't always use conventional commits or had a non-conventional commit slip in here or there for whatever reason (much better than refusing to run or something like that).

The **silent** part I'm not so sure about being a good thing, though - if commits are being skipped it seems like there is a good chance someone will know about that, even if it's expected and just to check on it.

Having managed to actually run into this now, I am convinced that #671 will fix this issue of `standard-version` silently ignoring signed commits when `log.showSignature` is enabled.

Whether or not parsing errors should be silently swallowed is another issue for another day.

dangreen:
#671 forces no-show-signature for parsing, so a user's Git configuration no longer affects it. Closing as fixed.

PR #671, opened 2020-09-10
fix(git-raw-commits): normalize git show signature option to false

thesmiley1:
This disables `git log` showing signatures (`--no-show-signature`)
unless `showSignature` is set to a truthy value.  This way differences
in developers' local configuration (`log.showSignature`) do not affect
tooling operation.  Callers of `gitRawCommits` will need to explicitly
request signatures (via `rawGitOpts`) if they are desired, instead of
the current behavior of the presence of signatures being undefined.

Fixes conventional-changelog/commitlint#2118


thesmiley1:
Maybe fixes #213?

I have not made any changes to tests, in part due to potential difficulty.  The current approach of dynamically creating a repository and commits would be complicated a great deal by the addition of gpg dependency/configuration, and trying to make that work on developers` machines as well as ci sounds... not fun.  I think a much simpler solution would be to change the current testing approach to use a static git repository with one or more already signed commits as a test fixture.

thesmiley1:
Additional thoughts:

[`GNUPGHOME`](https://www.gnupg.org/documentation/manuals/gnupg/GPG-Configuration.html#:~:text=locate%20the%20default%20home%20directory.-,GNUPGHOME,instead%20of%20%22~%2F.gnupg%22.) can be used to create temporary GPG configuration.  The GPG manual has a section specifically for [unattended key generation](https://www.gnupg.org/documentation//manuals/gnupg/Unattended-GPG-key-generation.html).

I'll wait for a response before doing anything further.

bcoe (approved):

ghiden:
Can we specify this option from `standard-version` command line?

thesmiley1:
> Can we specify this option from `standard-version` command line?

No.  This change does not add any additional options, nor is it affected by any existing options that users can set.

The change affects how commit data is fetched "internally" (via `git log` command), and the primary purpose is to prevent a bug from occurring (commit parser chokes on commit message) when developers have git's `log.showSignature` option enabled.

See conventional-changelog/commitlint#2118 for the type of bug this is intended to fix.

ghiden:
This actually breaks with old git versions.
```
> git log --no-show-signature
fatal: unrecognized argument: --no-show-signature
```

I happen to have a build server running git 2.7.4 and getting an error from `standard-version`.
I guess my only option would be update the old git then.

thesmiley1:
> This actually breaks with old git versions.
> 
> ```
> > git log --no-show-signature
> fatal: unrecognized argument: --no-show-signature
> ```
> 
> I happen to have a build server running git 2.7.4 and getting an error from `standard-version`.
> I guess my only option would be update the old git then.


I'm sorry to hear that. :slightly_frowning_face: 

It looks like the `log.showSignature` and `--no-show-signature` options were added in git/git@369dc4081c836bc17ee1debaf6688eb098359760, which was released in [v2.10.0](https://github.com/git/git/blob/master/Documentation/RelNotes/2.10.0.txt#:~:text=log.showSignature) (2016-09-02).

This project has a node support policy, but not a git support policy.  I'm not sure what, if any, support policy git has for itself. :thinking: 

If nothing else, it seems worth documenting in this project that `git >= 2.10` is required (in its current state).

If you think the matter warrants further attention, I would recommend opening a new issue (and linking this PR).  This PR has already been merged and any decisions regarding how to move forward would need to be made by project maintainers.

PR #672, opened 2020-09-15
feat(conventional-changelog-angular): Support for multiple feat and s…

Ricky-rick:
When we squash commits of a pull request we will keep the original commit message in the body of the new commit message.
This causes commitlint to fail because there is a second type and scope in the commit message.

**Example:**
A PR with the title 
"feat(pencil): add with and color option" 

That contains the commits
"feat(pencil): add with option" 
"feat(pencil): add color option" 

Wil result in the commit msg
"feat(pencil): add with and color option\n
feat(pencil): add with option\n 
"feat(pencil): add color option\n
Reviewed-on: <some PR info>"  

**Original regex:**
https://regex101.com/r/on6Ouy/2
*New regex*
https://regex101.com/r/on6Ouy/1

**workaround fix**
```js
// commitlint.config.js

module.exports = {
  parserPreset: {
    parserOpts: {
      headerPattern: /^(\w*)(?:\(([^:]*)\))?: (.*)$/,
    },
  },
}
```

### DIFF
--- a/packages/conventional-changelog-angular/parser-opts.js
+++ b/packages/conventional-changelog-angular/parser-opts.js
@@ -1,7 +1,7 @@
 'use strict'
 
 module.exports = {
-  headerPattern: /^(\w*)(?:\((.*)\))?: (.*)$/,
+  headerPattern: /^(\w*)(?:\(([^:]*)\))?: (.*)$/,
   headerCorrespondence: [
     'type',
     'scope',


bcoe:
@Ricky-rick would you be open to using the `conventional-commit` parser rather than the angular parser? we're working on formally defining the parser, vs., using a series of free-form regexes:

https://github.com/conventional-commits/parser

Going forward, this is the preset I'd like to put the most work into going forward.

PR #673, opened 2020-09-19
fix(git-raw-commits): silent false success when git is not present

thesmiley1:
Add callback function and error check to `execFile`.

Fixes conventional-changelog/commitlint#2136

### DIFF
--- a/packages/git-raw-commits/index.js
+++ b/packages/git-raw-commits/index.js
@@ -53,36 +53,29 @@ function gitRawCommits (rawGitOpts, rawExecOpts) {
     gitOpts.debug('Your git-log command is:\ngit ' + args.join(' '))
   }
 
-  let isError = false
-
   const child = execFile('git', args, {
     cwd: execOpts.cwd,
     maxBuffer: Infinity
+  }, function (err) {
+    if (err != null) {
+      readable.emit('error', err)
+    }
+
+    readable.emit('close')
   })
 
   child.stdout
     .pipe(split(DELIMITER + '\n'))
     .pipe(through(function (chunk, enc, cb) {
       readable.push(chunk)
-      isError = false
 
       cb()
     }, function (cb) {
       setImmediate(function () {
-        if (!isError) {
-          readable.push(null)
-          readable.emit('close')
-        }
+        readable.push(null)
 
         cb()
       })
-    }))
-
-  child.stderr
-    .pipe(through.obj(function (chunk) {
-      isError = true
-      readable.emit('error', new Error(chunk))
-      readable.emit('close')
     }))
 
   return readable
--- a/packages/git-raw-commits/test.js
+++ b/packages/git-raw-commits/test.js
@@ -29,8 +29,6 @@ describe('git-raw-commits', function () {
       })
       .pipe(through(function () {
         done('should error')
-      }, function () {
-        done('should error')
       }))
   })
 
@@ -191,5 +189,25 @@ describe('git-raw-commits', function () {
         expect(i).to.equal(3)
         done()
       }))
+  })
+
+  it('should emit an error if git is not available', function (done) {
+    try {
+      var path = process.env.PATH
+      process.env.PATH = ''
+
+      gitRawCommits()
+        .on('error', function (err) {
+          expect(err).to.be.ok // eslint-disable-line no-unused-expressions
+          done()
+        })
+        .pipe(through(function () {
+          done('should error')
+        }, function () {
+          done('should error')
+        }))
+    } finally {
+      process.env.PATH = path
+    }
   })
 })


packages/git-raw-commits/index.js
```diff
@@ -52,38 +52,31 @@ function gitRawCommits (rawGitOpts, rawExecOpts) {
     gitOpts.debug('Your git-log command is:\ngit ' + args.join(' '))
   }
 
-  var isError = false
```
thesmiley1, line 55: Variable removed because the only code where it changes is also removed.

packages/git-raw-commits/index.js
```diff
@@ -52,38 +52,31 @@ function gitRawCommits (rawGitOpts, rawExecOpts) {
...
-
   var child = execFile('git', args, {
     cwd: execOpts.cwd,
     maxBuffer: Infinity
+  }, function (err) {
+    if (err != null) {
+      readable.emit('error', err)
+    }
+
+    readable.emit('close')
```
thesmiley1, line 63: Error checking is changed from checking for output on STDERR to checking the error in a callback function.  `error` is emitted if a non-null error is passed to the callback, and `close` is always emitted.

packages/git-raw-commits/index.js
```diff
@@ -52,38 +52,31 @@ function gitRawCommits (rawGitOpts, rawExecOpts) {
...
   })
 
   child.stdout
     .pipe(split(DELIMITER + '\n'))
     .pipe(through(function (chunk, enc, cb) {
       readable.push(chunk)
-      isError = false
```
thesmiley1, line 66: Variable removed.

packages/git-raw-commits/index.js
```diff
@@ -52,38 +52,31 @@ function gitRawCommits (rawGitOpts, rawExecOpts) {
...
 
       cb()
     }, function (cb) {
       setImmediate(function () {
-        if (!isError) {
```
thesmiley1, line 71: Variable removed.

packages/git-raw-commits/index.js
```diff
@@ -52,38 +52,31 @@ function gitRawCommits (rawGitOpts, rawExecOpts) {
...
         cb()
       })
     }))
 
-  child.stderr
-    .pipe(through.obj(function (chunk) {
-      isError = true
-      readable.emit('error', new Error(chunk))
-      readable.emit('close')
-    }))
```
thesmiley1, line 85: The error value passed to `execFile` callback is checked instead of checking STDERR.

Is there a reason STDERR was checked instead?  If yes, this needs to stay and maybe a test should be added for whatever that use case is.

packages/git-raw-commits/test.js
```diff
@@ -29,8 +29,6 @@ describe('git-raw-commits', function () {
       })
       .pipe(through(function () {
         done('should error')
-      }, function () {
-        done('should error')
```
thesmiley1, line 33: The flush function _does_ get called before `error` is emitted now, so this is removed to not call `done` twice.

That seems okay... I couldn't think of a reason that should be a problem, and all tests are still passing.

packages/git-raw-commits/test.js
```diff
@@ -192,4 +190,24 @@ describe('git-raw-commits', function () {
         done()
       }))
   })
+
+  it('should emit an error if git is not available', function (done) {
+    try {
```
thesmiley1, line 195: Taking a second look at this, I think this `try`/`finally` _can_ be avoided with options to `rawExecOpts`... that would be much better.

packages/git-raw-commits/test.js
```diff
@@ -192,4 +190,20 @@ describe('git-raw-commits', function () {
...
+    rawExecOpts.env.PATH = ''
+
+    gitRawCommits({}, rawExecOpts)
+      .on('error', function (err) {
+        expect(err).to.be.ok // eslint-disable-line no-unused-expressions
+        done()
+      })
+      .pipe(through(function () {
+        done('should error')
+      }, function () {
```
thesmiley1, line 205: Here the flush function is _not_ called before `error` is emitted.  In this case there is no output whatsoever (empty or otherwise) so output processing is apparently skipped.

This is inconsistent with the above comment that the flush function is called before `error` is emitted, but I can't think of another case where this might happen.

packages/git-raw-commits/test.js
thesmiley1: Nope, I was right the first time...

The reason for the `try`/`finally` is to ensure that changes to the process environment are always reverted, even in case the test fails to complete for some reason, so as to avoid affecting other tests.  My first thought was to pass environment options via `rawExecOpts` (see 58d1871221533815879e0f552e395413c8b027a1), but that is not possible without further code changes because [`cwd` is the only property of `rawExecOpts` that gets passed to `execFile`](https://github.com/conventional-changelog/conventional-changelog/blob/be1246c68f5dc4e6f28996129951a75bbf1cf307/packages/git-raw-commits/index.js#L48-L60).  This code is ugly because I erred on the side of making the test more complicated vs. changing more critical code.

I think it makes the most sense for `normalizeExecOpts` to set the `maxBuffer` property and use that return value (that includes all properties of the object passed as `rawExecOpts` after being normalized) to pass to `execFile`, but I'm not sure that is in scope for fixing this bug.

escapedcat:
Thanks @thesmiley1 !

thesmiley1:
I squashed commits made when this PR was last worked on and rebased to fix conflicts. :slightly_smiling_face:

PR #656, opened 2020-06-18
feat: add support for '--skip-unstable' option

dudekaa:
Hi

during usage of your great tool in our team, we came with this little improvement that makes our lives easier.

During our development workflow we often create a few :smile: release candidates in process, which then makes changelog generation little tricky as their number vary.

This new _--skip-unstable_ simply skips all those (_alpha_, _beta_, _whatever_ too), so the default _--release-count=1_ works as expected.

Maybe this will be useful for some other too.

BR
Arnost

PR #674, opened 2020-09-24
feat(conventional-recommended-bump): support for '--skip-unstable'

admir86:
[This feature](https://github.com/conventional-changelog/conventional-changelog/pull/656) is exactly what I need. But I need it also in the conventional-recommended-bump module.
It should be possible to use the --skip-unstable option, of the git-semver-tags module.


admir86:
@bcoe any chance to merge and publish this feature in the next days ?

Issue #688, opened 2020-10-30
"git-semver-tags" don't read all tags


pablogm95:
When i run the command `git tag --sort=-"v:refname"` in my project I can list all my tags corretly:
```
2.4.0
2.3.0
2.2.0
2.1.0
2.0.1
2.0.0
1.0.0
0.1.0-beta.9.1
0.1.0-beta.9
0.1.0-beta.7
0.1.0-beta.6
0.1.0-beta.5
0.1.0-beta.4
0.1.0-beta.3
0.1.0-beta.2
0.1.0-beta.1
0.1.0-beta
```

But running `npx git-semver-tags` the result is:
```
2.0.0
1.0.0
0.1.0-beta.9.1
0.1.0-beta.9
0.1.0-beta.7
0.1.0-beta.6
0.1.0-beta.5
0.1.0-beta.4
0.1.0-beta.3
0.1.0-beta.2
0.1.0-beta.1
0.1.0-beta
```

This error produce that my CHANGELOG is not filled correctly because the latest tag is not recognized.

All my tags have been created in the same way

voidale:
same here

pablogm95:
I add a pull request to try to resolve this issue
https://github.com/conventional-changelog/conventional-changelog/pull/692

voidale:
found a way to fix this, you just need to tag the commits and it will work fine

pablogm95:
> found a way to fix this, you just need to tag the commits and it will work fine

Can you explain your solution plis?

mason-kanpai:
@pablogm95 

Run this command. This is what **git-semver-tags** sees as tags
```
git log --decorate |grep tag
```

To fix this, you need to tag the commits again. In your case, please find your commit SHA for 2.4.0
```
git tag -a v2.4.0 commit_SHA -m "chore(release): 2.4.0"
```

https://devconnected.com/how-to-create-git-tags/#Create_Git_Tag_for_Commit

dangreen:
Tag discovery now accepts all: true and walks all refs. Configure tags: { all: true } (or call getSemverTags({ all: true })) for tags outside the current branch history. Closing as completed.

PR #693, opened 2020-11-02
Fix git semver tags recognize all tags

pablogm95:
This pull request resolves the https://github.com/conventional-changelog/conventional-changelog/issues/688 issue

### DIFF
--- a/packages/git-semver-tags/index.js
+++ b/packages/git-semver-tags/index.js
@@ -4,7 +4,7 @@ var proc = require('process')
 var exec = require('child_process').exec
 var semverValid = require('semver').valid
 var regex = /tag:\s*(.+?)[,)]/gi
-var cmd = 'git log --decorate --no-color'
+var cmd = 'git tag --sort=-"v:refname"'
 var unstableTagTest = /.*-\w*\.\d$/
 
 function lernaTag (tag, pkg) {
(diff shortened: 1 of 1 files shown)


voidale:
Good job,  git tag --sort=-"v:refname" list all tags properly! for me as well!

bcoe:
I think the issue with this approach would be that it returns tags across multiple branches, where as some of the libraries that rely on `git-semver-tags` assume that it's only returning tags that are associated with the current branch that it's running against.

_This is at least my guess, as to why it was implemented in such a way._

bcoe:
@pablogm95 what if you made this an option you could pass in, so that people could opt in to listing all tags regardless of branch? or, alternatively, is there a way to tell the `tag` command to only run against the current branch?

pablogm95:
Hello @bcoe . Your suggestion sounds good. I'll work on it these days.

joshuasimon-taulia:
> I think the issue with this approach would be that it returns tags across multiple branches, where as some of the libraries that rely on `git-semver-tags` assume that it's only returning tags that are associated with the current branch that it's running against.
> 
> _This is at least my guess, as to why it was implemented in such a way._

is this documented anywhere? our CI process pushes an additional commit to the tag on release which causes new tags to go undetected

PR #708, opened 2020-12-21
feat(conventionalcommits): allow the scope to be omitted

felipecrs:
Example:

```javascript
{ type: 'build', scope: 'deps', section: 'Dependencies', omitScope: true }
```

produces:

```markdown
## Dependencies

- upgrade example from 1 to 2
```

rather than:

```markdown
## Dependencies

- **deps:** upgrade example from 1 to 2
```


bcoe:
@felipecrs I'm wondering if we could avoid adding an additional configuration option, what if you passed the commits through a transform stream and remove the scope:

```
class PostProcessCommits extends Transform {
  _transform(
    chunk: ParsedConventionalCommit,
    _encoding: string,
    done: Function
  ) {
  delete chunk.scope
}
```

☝️ something like this.

felipecrs:
I'm not sure if I understood. Apart from the implementation suggestion, are you saying that we should automatically omit the scope if the user defined a scope for the type?

bcoe:
>  Apart from the implementation suggestion, are you saying that we should automatically omit the scope if the user defined a scope for the type?

@felipecrs because there are so many configuration options that someone could set, I'm hesitant to add a toggle; an alternative option would be changing the commit object before it goes to the parser, or passing an alternate template for commits.

felipecrs:
> an alternative option would be changing the commit object before it goes to the parser, or passing an alternate template for commits.

Okay... I'm just confused, how could I set this with https://github.com/semantic-release/release-notes-generator then?

bcoe:
@felipecrs I'll make an effort to pull together an example 👍

felipecrs:
@bcoe ping :)

maximenathan:
Following :)

felipecrs:
Hello, could a maintainer take a look at this? I believe it's ready to be reviewed, if not to be merged.

PR #711, opened 2020-12-25
feat: add helper for parsing array of commits

bcoe:
## What is this?

This PR adds a method that allows an array of commit objects to be passed to the writer, rather than piping a stream.

## Why is this?

* The streaming API can be burdensome to use when a user of the API has access to an already parsed array of commits.
* We've [noticed bugs related to stream order](https://github.com/googleapis/release-please/pull/581) in some Node.js versions.
* We're currently working on a [formal parser for "Conventional Commits"](https://github.com/conventional-commits/parser), which does not expose a streaming API.

### DIFF
--- a/packages/conventional-changelog-writer/index.js
+++ b/packages/conventional-changelog-writer/index.js
@@ -193,7 +193,6 @@ conventionalChangelogWriterParseStream.parseArray = (_rawCommits, _context, _opt
   }
   if (options.reverse) {
     entries.reverse()
-    console.info('we reverse')
     return util.generate(options, commits, context, savedKeyCommit) + entries.join('')
   } else {
     return entries.join('') + util.generate(options, commits, context, savedKeyCommit)
--- a/packages/conventional-changelog-writer/index.js
+++ b/packages/conventional-changelog-writer/index.js
@@ -80,8 +80,9 @@ function conventionalChangelogWriterInit (context, options) {
   return { context, options, generateOn }
 }
 
-function conventionalChangelogWriterParseStream (_context, _options) {
-  const { context, options, generateOn } = conventionalChangelogWriterInit(_context, _options)
+function conventionalChangelogWriterParseStream (context, options) {
+  let generateOn
+  ({ context, options, generateOn } = conventionalChangelogWriterInit(context, options))
   let commits = []
   let neverGenerated = true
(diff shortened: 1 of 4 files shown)


packages/conventional-changelog-writer/index.js
```diff
@@ -163,4 +167,37 @@ function conventionalChangelogWriter (context, options) {
...
+      savedKeyCommit = keyCommit
+      commits = []
+    }
+    if (commit) {
+      commits.push(commit)
+    }
+  }
+  if (options.reverse) {
+    entries.reverse()
+    console.info('we reverse')
```
jbottigliero, line 196: It looks like this `console.info` might have snuck into the PR.

packages/conventional-changelog-writer/test/index.spec.js
```diff
@@ -149,15 +158,17 @@ describe('conventionalChangelogWriter', function () {
...
       let i = 0
-
+      const context = {
+        version: '0.5.0',
+        title: 'this is a title',
+        host: 'https://github.com',
+        repository: 'a/b',
+        linkReferences: false
+      }
+      const changelog = conventionalChangelogWriter.parseArray(commits, {})
```
jbottigliero, line 168: This call/test for`.parseArray` looks slightly different than the stream; Should the `context` be passed so they match?

jbottigliero (approved):
This looks great to me! Caught a few minor things in review (see inline comments). 

It seems like there might be additional abstraction/sharing that could occur in some of the tests (ie. `context` and `option` declaration sharing), but that doesn't seem super necessary and potentially just cumbersome with the intent of moving over to the parser eventually.

nit/question: I've only seen the leading`_` used for unused arguments or variable declarations. In most of the additions here it's being used to allow "reuse" of the argument name as a variable – seems fine to me, but thought it was worth mentioning.

bcoe:
> nit/question: I've only seen the leading_ used for unused arguments or variable declarations. In most of the additions here it's being used to allow "reuse" of the argument name as a variable – seems fine to me, but thought it was worth mentioning.

@jbottigliero I think you're right that this approach was unclear, I switched to using "assignment without declaration" -- which apparently works as long as you enclose the destructuring assignment in parenthesis.

PR #737, opened 2021-01-08
fix: align lodash dependency across packages

armano2:
https://lodash.com/per-method-packages


bcoe (approved):

bcoe:
@armano2 thank you for the contribution, I'd noticed that lodash was attempting to move folks away from the individual modules.

PR #743, opened 2021-01-17
fix: migrate lodash.assign to native Object.assign

armano2:
Migrates use of lodash assign to native Object.assign, additionally i changed some imports as we don't have to import entire lodash package when we are using only one function from it

### DIFF
--- a/packages/conventional-changelog-core/lib/merge-config.js
+++ b/packages/conventional-changelog-core/lib/merge-config.js
@@ -180,7 +180,11 @@ function mergeConfig (options, context, gitRawCommitsOpts, parserOpts, writerOpt
           }
           context.owner = context.owner || repo.user || ''
           context.repository = context.repository || repo.project
-          context.repoUrl = browse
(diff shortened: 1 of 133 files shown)


jimmywarting (approved):

packages/conventional-changelog-core/lib/merge-config.js
```diff
@@ -221,7 +221,7 @@ function mergeConfig (options, context, gitRawCommitsOpts, parserOpts, writerOpt
         if (type) {
           hostOpts = require('../hosts/' + type)
 
-          context = _.assign({
+          context = Object.assign({
             issue: hostOpts.issue,
             commit: hostOpts.commit
           }, context)
```
jimmywarting, line 227:
```js
          context = {
            issue: hostOpts.issue,
            commit: hostOpts.commit,
            ...context
          }
```

jimmywarting (approved):

PR #745, opened 2021-01-17
refactor(gulp-conventional-changelog): remove unnecessary object-assign polyfill

armano2:
https://github.com/sindresorhus/object-assign

> Node.js 4 and up, as well as every evergreen browser (Chrome, Edge, Firefox, Opera, Safari), support Object.assign() 🎉. If you target only those environments, then by all means, use Object.assign() instead of this package.

#753


jimmywarting (approved):

packages/gulp-conventional-changelog/index.js
```diff
@@ -1,15 +1,14 @@
...
-const assign = require('object-assign')
 const concat = require('concat-stream')
 const conventionalChangelog = require('conventional-changelog')
 const PluginError = require('plugin-error')
 const fancyLog = require('fancy-log')
 const through = require('through2')
 
 module.exports = function (opts, context, gitRawCommitsOpts, parserOpts, writerOpts) {
-  opts = assign({
+  opts = Object.assign({
```
jimmywarting, line 11: How about just using object destructing instead. vscode have better understanding of it's properties that way...

```js
opts = {
    // TODO: remove this when gulp get's a real logger with levels
    verbose: process.argv.indexOf('--verbose') !== -1,
    ...opts
}
```

armano2: i do agree that `object destructing` has a better syntax, but sadly we are not able to use to as this code has to work on node 10

jimmywarting (approved):

armano2:
solved by #959

PR #747, opened 2021-01-19
refactor(filter): use lodash/mapValues instead of modify-values

armano2:
https://lodash.com/docs/4.17.15#mapValues

#753


armano2:
solved by #959

PR #749, opened 2021-01-19
refactor(standard-changelog): replace sprintf-js with util.format

armano2:
https://nodejs.org/docs/latest/api/util.html#util_util_format_format_args

available since node v0.5.3

### DIFF
--- a/packages/standard-changelog/index.js
+++ b/packages/standard-changelog/index.js
@@ -6,7 +6,7 @@ const fs = require('fs')
 const accessSync = require('fs-access').sync
 const chalk = require('chalk')
 const figures = require('figures')
-const sprintf = require('sprintf-js').sprintf
+const format = require('util').format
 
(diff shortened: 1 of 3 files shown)


packages/standard-changelog/index.js
```diff
@@ -26,7 +26,7 @@ conventionalChangelog.createIfMissing = function (infile) {
 }
 
 conventionalChangelog.checkpoint = function (msg, args) {
-  console.info(chalk.green(figures.tick) + ' ' + sprintf(msg, args.map(function (arg) {
+  console.info(chalk.green(figures.tick) + ' ' + format(msg, ...args.map(function (arg) {
```
armano2, line 29: alterantively this can be done without format as console.info support sprintf dirrectly
```suggestion
  console.info(chalk.green(figures.tick) + ' ' + msg, ...args.map(function (arg) {
```

dangreen:
done in https://github.com/conventional-changelog/conventional-changelog/pull/1111

PR #750, opened 2021-01-21
chore: remove no longer needed pinkie-promise polyfill from tests

armano2:
Remove no longer needed promise polyfill
#753


armano2:
removed in #1067

Issue #753, opened 2021-01-22
conventional-changelog and its dependencies


armano2:
### General Informaitions

Recently i started investigating and making small pull requests to reduce/simplify dependency graph for packages from this repository

- importing `conventional-changelog` requires you to download 157 dependencies
  https://npm.anvaka.com/#/view/2d/conventional-changelog
- importing `git-raw-commints` requires you to download 78 dependencies
  https://npm.anvaka.com/#/view/2d/git-raw-commits
- importing just a `parser` requires you to download 89
  https://npm.anvaka.com/#/view/2d/conventional-commits-parser

as reference:
- @babel/core  - 40 dependencies - https://npm.anvaka.com/#/view/2d/%2540babel%252Fcore
- @babel/cli - 22 dependencies - https://npm.anvaka.com/#/view/2d/%2540babel%252Fcli
- yargs - 16 dependencies - https://npm.anvaka.com/#/view/2d/yargs
- meow - 66 dependencies -  https://npm.anvaka.com/#/view/2d/meow
- mocha - 96 dependencies -  https://npm.anvaka.com/#/view/2d/mocha
- eslint - 108 dependencies - https://npm.anvaka.com/#/view/2d/eslint

https://snyk.io/blog/how-to-maintain-npm-dependencies-in-your-project/

------------------------------

### Recommended refactoring:
split cli logic from packages, as consumers of your packages does not require this functionality at all

example:
introduce new package `git-raw-commits-api` and move index.js from `git-raw-commits`
`git-raw-commits` should depend on `git-raw-commits-api` and can rexeport all functionality to index for backward compatibility, that should remove requirement of installing cli for packages from within this project or external consumers
- `lerna`
- `commitlint`
- `git-commits-since`
- `autorelease`
- `conventional-git-log`
- `semantic-release-github`

https://www.npmjs.com/browse/depended/git-raw-commits

in case of this package this split changes number of used packages from 78 to 9

This should be applied to all packages that provide cli as other libraries does not require cli at all

note: name is just an example

------------------------------

### Progress so far

Some smaller changes that should remove deprecated, not maintained or unnecessary dependencies:
- [x] #742 - `q` - it was a great package when promises where not available **[not maintained]**
- [ ] #750 - `pinkie-promise` - polyfill for Promise [not maintained]
- [x] #745 - `object-assign` - polyfill for Object.assign that is needed for node less than 4 **[not maintained]**
[...]

PR #759, opened 2021-01-31
refactor: replace lodash/isString with typeof check

armano2:
#753

### DIFF
--- a/packages/conventional-changelog-writer/lib/util.js
+++ b/packages/conventional-changelog-writer/lib/util.js
@@ -12,20 +12,20 @@ function compileTemplates (templates) {
   const footerPartial = templates.footerPartial
   const partials = templates.partials
 
-  if (_.isString(headerPartial)) {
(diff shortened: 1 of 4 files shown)


packages/conventional-changelog-writer/lib/util.js
```diff
@@ -12,20 +12,20 @@ function compileTemplates (templates) {
...
     Handlebars.registerPartial('header', headerPartial)
   }
 
-  if (_.isString(commitPartial)) {
+  if (typeof commitPartial === 'string') {
     Handlebars.registerPartial('commit', commitPartial)
   }
 
-  if (_.isString(footerPartial)) {
+  if (typeof footerPartial === 'string') {
```
bcoe, line 23: Same comment as `typeof boolean`:

```
> _.isString(new String('hey'))
true
```

It might be worth making your own helper that checks both `typeof` and `instanceof`.

bcoe (changes requested):

PR #760, opened 2021-01-31
refactor: replace lodash/isBoolean with typeof check

armano2:
#753

### DIFF
--- a/packages/conventional-changelog-writer/index.js
+++ b/packages/conventional-changelog-writer/index.js
@@ -15,7 +15,7 @@ function conventionalChangelogWriterInit (context, options) {
     date: dateFormat(new Date(), 'yyyy-mm-dd', true)
   }, context)
 
-  if (!_.isBoolean(context.linkReferences) && (context.repository || context.repoUrl) && context.commit && context.issue) {
+  if (typeof context.linkReferences !== 'boolean' && (context.repository || context.repoUrl) && context.commit && context.issue) {
     context.linkReferences = true
   }
 
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,7 @@
 /lerna-debug.log
 package-lock.json
 
+# Editors
+.idea
+.vscode
+.DS_Store
--- a/packages/conventional-changelog-core/lib/merge-config.js
+++ b/packages/conventional-changelog-core/lib/merge-config.js
@@ -202,7 +202,7 @@ function mergeConfig (options, context, gitRawCommitsOpts, parserOpts, writerOpt
         }
       }
 
-      if (!_.isBoolean(options.outputUnreleased)) {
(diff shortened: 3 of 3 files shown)


packages/conventional-changelog-writer/index.js
```diff
@@ -15,7 +15,7 @@ function conventionalChangelogWriterInit (context, options) {
     date: dateFormat(new Date(), 'yyyy-mm-dd', true)
   }, context)
 
-  if (!_.isBoolean(context.linkReferences) && (context.repository || context.repoUrl) && context.commit && context.issue) {
+  if (typeof context.linkReferences !== 'boolean' && (context.repository || context.repoUrl) && context.commit && context.issue) {
```
bcoe, line 18: If someone was providing `new Boolean(true)` or `new Boolean(false)` as an option, I think `_.isBoolean` would indicate that it's boolean, whereas typeof would not.

Perhaps do:

```
!(typeof context.linkReferences === 'boolean' || context.linkReferences instanceof Boolean)
```

bcoe (changes requested):

packages/conventional-changelog-writer/index.js
armano2: `new Boolean(true)`  and `new Boolean(false)` is always "truthly" 

```ts
Boolean(new Boolean(true)) // true
Boolean(new Boolean(false)) // true
```

than there is no contextual difference between them, witch means is still correct
```ts
context.linkReferences = true
```

wesleytodd: I don't typically test values like this, I find that it is pointless for most programs.  In this case, why not just use any not-`undefined` value (`typeof context.linkReferences !== 'undefined'`) as indicating the user passed an option and it should not use the default? So what if someone goes out of their way to create an instance of `Boolean`? What situation are we actually protecting users from that is worth keeping the regular cve's filed against `lodash` in their audits (I read this PR as having the goal of removing `lodash` entirely, if that is not the goal then I don't see removing these calls as worth it)?

But, if you are going to change it why not just test explicitly against `true` and `false`?

armano2: > I read this PR as having the goal of removing lodash entirely

thats a general goal, as there is no really need for it here, but doing this is going to requireway more changes -> generally i preffer to create smaller PR's that have smaller impact.

i'm not against closing this and keeping using lodash for now

bcoe: > I'm not against closing this and keeping using lodash for now.

I think I'd personally rather see a bigger PR that removes lodash, if we think that's worthwhile (and we're not actually leveraging any of its more valuable features).

Then we could take that PR as a major, and I'd be less concerned about this minor differences in behavior.

PR #769, opened 2021-02-17
test: update tmp folder and drop shelljs

armano2:
This change drops shelljs from tests in all test packages package, and moves tests tmp folder to system tmp folder

PR #770, opened 2021-02-17
test(conventional-changelog-atom): update tmp folder and drop shelljs

armano2:
This change drops shelljs from tests in `conventional-changelog-atom` package, and moves tests tmp folder to system tmp folder


armano2:
suppressed by #769

PR #774, opened 2021-02-26
fix: optimise `gitSemverTags`

kellyselden:
This resulted in speed bonuses on our massive monorepo.

### DIFF
--- a/packages/git-semver-tags/test.js
+++ b/packages/git-semver-tags/test.js
@@ -61,7 +61,7 @@ describe('git-semver-tags', function () {
 
     gitSemverTags(function (err, tags) {
       if (err) done(err)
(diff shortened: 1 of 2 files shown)


armano2 (approved):

packages/git-semver-tags/test.js
```diff
@@ -90,29 +88,35 @@ describe('git-semver-tags', function () {
   })
 
   it('should be in reverse chronological order', function (done) {
-    writeFileSync('test4', '')
-    shell.exec('git add --all && git commit -m"Fourth commit"')
-    shell.exec('git tag v1.0.0')
+    // new method is too fast
+    setTimeout(function () {
```
armano2, line 92: wdym to fast?

kellyselden: closing for now because I can't get the tests to pass reliably.

Issue #612, opened 2020-03-18
conventional-recommended-bump: always takes first tag returned by git-semver-tags, while it may not be the highest version available


micxxx:
For repositories leveraging Git Flow, it may happen that the latest tag pushed to main branch is not the highest version of the package available: release/ branches for previous major versions may receive bugfixes, and require releases. In Git Flow, those bugfixes and tags would automatically be merged to development branch.

In result, one's tag history may look like:
2.5.2
3.0.0
2.5.1
2.5.0
etc.

**Current implementation of conventional-recommended-bump always takes the first result of git-semver-tags**: https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-recommended-bump/index.js#L74

Please correct me if I'm wrong, **but the highest version returned by git-semver-tags should be used**.

dangreen:
Thanks for the report. The recommended bump is intentionally based on the most recent reachable semver tag, rather than the numerically highest tag across all branches. In a maintenance-branch topology, the global highest version can be on another branch and yield the wrong commit range. Use --from <tag-or-sha> (or the API equivalent) when a different baseline is needed. Closing as not planned.

Issue #194, opened 2017-06-03
Allow passing in version parameter from command line


Alxandr:
I'm using `conventional-changelog-cli` to release something that's not a node repository (that still requires me to have node installed for other reasons). Currently I have as part of my "release" script that it generates a `package.json` with the correct version in it, calls `conventional-changelog`, then just deletes the `package.json` again. It would be great if parameters from `package.json` could be specified other ways. At least the version parameter should be able to set using the command line easily IMHO.

stevemao:
I thought about it, and it was an option in the old version. But it doesn't fit the paradigm since version is just a context variable. I could argue things like [repository](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-writer#repository) should be able to specified as a flag too. Use [config](https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-cli/cli.js#L45) flag, It's a generic solution.

Xapphire13:
Is there any documentation on the config script format? The linked example is super obscure

hutson:
I came across a similar issue. I needed to generate release notes for the latest git tag (the package had already been released and I was just generating notes for the previous release). Also of note, I don't update the version in `package.json`, so the `version` field always says `1.0.0`.

What I ended up doing was creating a context file with the version I wanted generate release notes for, and then telling `conventional-changelog` CLI to generate release notes for the latest two versions (a release count of 1 always generates release notes for commits since the last tag).

Here's the script I used:

```bash
echo "{\"version\": \"${CI_COMMIT_TAG}\"}" > /tmp/context.json && $(yarn bin)/conventional-changelog --preset @hutson/config --release-count 2 --context /tmp/context.json
```

ghostsquad:
Documentation on how to use the `--context` besides the example provided by @hutson would be great too.

dangreen:
For a non-Node repository, pass the version via context JSON and the corresponding CLI option. Closing as answered.

PR #777, opened 2021-03-09
(conventional-recommended-bump) Specify base tag to analyze tags from

kf6kjg:
This feature allows the user to specify the base tag from which to scan for version tags, allowing for tooling like standard-version to pull the previous version from some other location than the most recent tag. As simple icing I added the flag to the both the recommended-bump and changelog CLIs so that users of those tools can enjoy this feature there as well.

I chose not to attempt a "to" flag as that would make this much more complex and is not needed for my use case just now.

Also [fixes a bug in conventional-recommended-bump](https://github.com/conventional-changelog/conventional-changelog/commit/6644339e81c39a73df414e57a7c5d51e72e6cbe4) that would cause a callback to not be called if the `gitRawCommits` function emitted an error instead of returning a stream.

AFAICT this could be a fix for both #194 and #612, but this was mainly to solve a problem in standard-version.  I'll link the PR to that repo once I've got it created.

In my use-case I don't store the changelog in the committed code as it has problems when dealing with branches.  I don't see any generic way to solve that problem, so instead I use my own repo-specific algorithm to determine which tag in the history is the "correct" parent tag depending on several bits of repo-specific context and then need to generate the changelog and recommended bump based off that tag, and not before.

### DIFF
--- a/packages/conventional-changelog-cli/cli.js
+++ b/packages/conventional-changelog-cli/cli.js
@@ -50,6 +50,7 @@ const cli = meow(`
       -c, --context             A filepath of a json that is used to define template variables
       -l, --lerna-package       Generate a changelog for a specific lerna package (:pkg-name@1.0.0)
       -t, --tag-prefix          Tag prefix to consider when reading the tags
+      --tag-from                Tag to start reading from when reading the tags
       --commit-path             Generate a changelog scoped to a specific directory
 `, {
   booleanDefault: undefined,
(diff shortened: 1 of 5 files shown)


dangreen:
The current changelog and recommended-bump CLIs both support --from <tag-or-sha> (and --to) for selecting the commit range. This supersedes the proposed --tag-from implementation. Closing as completed.

PR #787, opened 2021-04-08
fix(conventionalcommits): always use h2

Mesteery:
Now H2 is used all the time whether it is a patch or not.
Before an H3 was used when it was a patch, but this is bad, because what is supposed to be the title is of the same level as its content (the changelogs sections), so the title (when it was a patch) was hardly differentiable from its content (same title level: H3).
Moreover it could avoid bugging tools that are based on the title level.

### DIFF
--- a/packages/conventional-changelog-conventionalcommits/templates/header.hbs
+++ b/packages/conventional-changelog-conventionalcommits/templates/header.hbs
@@ -1,13 +1,9 @@
-{{#if isPatch~}}
-  ###
-{{~else~}}
-  ##
-{{~/if}} {{#if @root.linkCompare~}}
-  [{{version}}]({{compareUrlFormat}})
-{{~else}}
-  {{~version}}
-{{~/if}}
-{{~#if title}} "{{title}}"
-{{~/if}}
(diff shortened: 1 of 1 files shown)


bcoe:
@Mesteery this would be a breaking change. Perhaps you could instead pass in a custom template for your use case?

Mesteery:
Yes and no. It's for release-please.
Personally I think that the level of the titles should be well respected: the main title should always be superior to the others, never on the same level.

bcoe:
@Mesteery funny enough, I think for release-please we pass our own templates, so this wouldn't actually correct the template in its case.

Mesteery:
https://github.com/googleapis/release-please/blob/37f4ec236be3bd99e74162a0cc16acd47042528f/src/conventional-commits.ts#L206 is using `conventional-changelog-conventionalcommits` (except for Ruby releaser who use custom templates.

dangreen:
Thanks. This behavior was implemented in #920: version headers now always use H2, and the current renderer continues to do so. This PR is therefore superseded. Closing as completed.

PR #809, opened 2021-06-03
Update get-pkg-repo module

ryan-st:
Update get-pkg-repo module to avoid ddos vulnerability in downstream dependency trim-newlines V1.0.0.

Reference: https://snyk.io/vuln/SNYK-JS-TRIMNEWLINES-1298042

Dependency tree when running `yarn why trim-newlines`:

`=> Found "get-pkg-repo#trim-newlines@1.0.0"`
`info Reasons this module exists`
`- "standard-version#conventional-changelog#conventional-changelog-core#get-pkg-repo#meow" depends on it`
`- Hoisted from "standard-version#conventional-changelog#conventional-changelog-core#get-pkg-repo#meow#trim-newlines"`

This change fixes the vulnerability described above.

### DIFF
--- a/packages/conventional-changelog-core/package.json
+++ b/packages/conventional-changelog-core/package.json
@@ -30,7 +30,7 @@
     "conventional-changelog-writer": "^5.0.0",
     "conventional-commits-parser": "^3.2.0",
     "dateformat": "^3.0.0",
-    "get-pkg-repo": "^1.0.0",
+    "get-pkg-repo": "^4.0.0",
     "git-raw-commits": "^2.0.8",
     "git-remote-origin-url": "^2.0.0",
     "git-semver-tags": "^4.1.1",


victoria-miltcheva (approved):

jimmyhlee94 (approved):

TheRealWaldo (approved):

nicojs (approved):

shamilpatel25 (approved):

Russ-Corsha (approved):

Russ-Corsha:
This fixes `npm audit` for this repo (and all of its upstream dependencies)

samgregory88 (approved):

dahukish:
There is a test breakage here `packages/conventional-changelog-core/test/test.js:340` as the `browse` function now returns `undefined` for the missing parts of the `Url`.

I added `context.repoUrl = /undefined/i.exec(browse) ? context.host : browse` to `packages/conventional-changelog-core/lib/merge-config.js:183` which fixes it (for me locally).

stevemao:
@ryan-st are you able to fix what @dahukish has said?

stevemao:
The `undefined` in regex seems strange. but if you can find a better way that'd be great!

bcoe:
@ryan-st @stevemao I ended up landing this fix last night in another PR ... the behavior of `get-pkg-repo`@`4.x.x` seems a bit odd, rather than checking for an `undefined` string in the URL, I check whether we actually found an `org` and `repo`.

This still seems slightly weird though, it might be better to just use GitHub URL detection, and then let users implement other types of URL formats themselves using templates.

Any ways, tests are passing now and the security issue should be patched. `get-pkg-repo` is also now automated like this repo, so we can land patches easier.

PR #836, opened 2021-08-25
chore(deps): use @rubendew/get-pkg-repo

githrdw:
Several people have been waiting for an update of `get-pkg-repo` since June. This [update](https://github.com/conventional-changelog/get-pkg-repo/pull/62) fixes a [security vulnerability](https://github.com/conventional-changelog/get-pkg-repo/issues/60), but it hangs on a [permissions issue](https://github.com/conventional-changelog/get-pkg-repo/pull/66) when publishing to npm. 
So, it may not be the most elegant solution and may not be desirable, but I would like to offer the package `get-pkg-repo` under my own npm account.
https://www.npmjs.com/package/@rubendew/get-pkg-repo

https://github.com/conventional-changelog/standard-version/issues/772

### DIFF
--- a/packages/conventional-changelog-core/package.json
+++ b/packages/conventional-changelog-core/package.json
@@ -30,7 +30,7 @@
     "conventional-changelog-writer": "^5.0.0",
     "conventional-commits-parser": "^3.2.0",
     "dateformat": "^3.0.0",
-    "get-pkg-repo": "^4.0.0",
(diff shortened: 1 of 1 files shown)


bcoe:
The ability to publish was unblocked:

https://github.com/conventional-changelog/get-pkg-repo/blob/4.x/CHANGELOG.md

PR #842, opened 2021-09-08
Change to use trim-newline because trim-off-newline has CVE-2021-23425 issue

fang-lin:


bcoe:
@fang-lin thank you for the patch, this is a duplicate of https://github.com/conventional-changelog/conventional-changelog/pull/841, which I'm going to merge.

PR #866, opened 2021-10-30
docs(conventionalcommits): fix template examples

ingmarh:
Placeholders are denoted with two curly braces.

The examples were helpful for a quick start, but I was wondering why the template was not expanded. Needed to look in the source code to find out why and then it was obvious.


bcoe:
(_landed this as `fix` because I'm in the middle of testing some updates to release tooling_).

Thanks for the contribution @ingmarh.

PR #868, opened 2021-11-02
fix(conventional-changelog-conventionalcommits)!: use `## <small>` for patch versions

### LINKED ISSUE #867: conventional-changelog-conventionalcommits: Patch version headings shouldn’t be a different size
Same as #214, but for conventional-changelog-conventionalcommits. Heading levels indicate a structure. Patch versions should use `## <small>` instead of `### ` , as @marcysutton pointed out.

shivjm:
This is a backwards-incompatible change, so I would understand if it needed to be behind a configuration flag or something. I’ll leave it open pending feedback (or merging, of course!).

Closes #867

EDIT:
I got the tests running but they universally fail even on a clean upstream clone inside a Docker container, so I don’t think they’ll be much help here.

### DIFF
--- a/packages/conventional-changelog-conventionalcommits/templates/header.hbs
+++ b/packages/conventional-changelog-conventionalcommits/templates/header.hbs
@@ -1,8 +1,6 @@
-{{#if isPatch~}}
-  ###
-{{~else~}}
-  ##
-{{~/if}} {{#if @root.linkCompare~}}
+## {{#if isPatch~}}
+<small>
+{{~/if}}{{#if @root.linkCompare~}}
   [{{version}}]({{compareUrlFormat}})
 {{~else}}
   {{~version}}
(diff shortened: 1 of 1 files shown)


languitar:
Would it also be an option to go in this direction regarding my comment at https://github.com/conventional-changelog/conventional-changelog/issues/867#issuecomment-1021145863 ?

```patch
diff --git a/packages/conventional-changelog-conventionalcommits/templates/header.hbs b/packages/conventional-changelog-conventionalcommits/templates/header.hbs
index a4d6ec4..d7efc8f 100644
--- a/packages/conventional-changelog-conventionalcommits/templates/header.hbs
+++ b/packages/conventional-changelog-conventionalcommits/templates/header.hbs
@@ -1,8 +1,4 @@
-{{#if isPatch~}}
-  ###
-{{~else~}}
-  ##
-{{~/if}} {{#if @root.linkCompare~}}
+## {{#if @root.linkCompare~}}
   [{{version}}]({{compareUrlFormat}})
 {{~else}}
   {{~version}}
```

shivjm:
Sure. I used `small` to preserve the appearance like in #214, but since the element has its own semantics I’d be happier leaving it out.

shivjm:
I guess #920 supersedes this.

Issue #102, opened 2015-09-08
use full hash


oupala:
I'm using conventional-changelog to write a changelog in markdown.

I'm also using wrtier-opts to change the generated url for each commit. My forge is using the full hash in the url to reference each commit.

I managed to alter the url, but I really need the full hash to generate the url.

Note that I would still need the truncated hash (7 chars) so that the visible label of the link is not too long, but the full hash would be used in the url.

Here is an example of a generated url:

```
* feat(): add changelog page ([d11e82b](https://www.myforge.org/plugins/git/owner/project?p=project.git&a=commit&h=d11e82b))
```

that make the following template (see `shortHash` and `fullHash`):

```
* feat(): add changelog page ([<shortHash>](https://www.myforge.org/plugins/git/owner/project?p=project.git&a=commit&h=<fullHash>))
```

Is it possible to achieve this need?

stevemao:
change https://github.com/ajoslin/conventional-changelog/blob/master/presets/angular.js#L40-L42 in your own settings.

oupala:
Thanks @stevemao for pointing me the right file to change the hash computation.

In order to achieve what I needed, I had to change this file https://github.com/ajoslin/conventional-changelog/blob/master/presets/angular.js#L40-L42 the following way:

```js
      if (typeof commit.hash === 'string') {
        commit.shortHash = commit.hash.substring(0, 7);
      }
```

It is still possible to use the old (long) hash value, but I can also use the new shortHash.

My problem is that I don't see any easy way to add a **preset file** as a parameter of the command line. Hence I had to modify the original file from the original `conventional-changelog` which is installed globally on my workstation. This is not easily portable into continuous integration.

Three questions:

- Do you think this change can be added to the original Angular preset file? (and other preset as well?)
- Or can we imagine to pass a preset file as a parameter of the command line?
- In the end, hash lenght is more tightly linked to presentation (the writer) than to the preset. Would it be a better place to move the hash computation later in the process? Maybe somewhere in the writer opts?

stevemao:
Unfortunately you have do something like this
```
conventional-changelog -p angular -i CHANGELOG.md -w --writer-opts your-writer-opts-file.js
```

stevemao:
> It is still possible to use the old (long) hash value, but I can also use the new shortHash.
```js
if (typeof commit.hash === 'string') {
    commit.shortHash = commit.hash.substring(0, 7);
}
```
Then change your template?

[...]

PR #880, opened 2021-12-06
feat: full hash in commit.hbs for links

mrt181:
This changes the link to a commit to use the full hash instead of the shortened hash that is used in the link preview.
see also: https://github.com/conventional-changelog/conventional-changelog/issues/102#issuecomment-964914871


bcoe:
@mrt181 is the reason for this that the full has does not work on some platforms?

mrt181:
Yes, the short hash does not work with bitbucket and azure DevOps for example

bcoe:
@mrt181 this seems like a good fix 👌 but mind updating tests?

mrt181:
> @mrt181 this seems like a good fix 👌 but mind updating tests?

Hey, how can I run/ debug them in vs code? mocha sidebar doesn't seem to work?

mrt181:
@bcoe Those tests fail even when i revert my changes?

armano2:
> Yes, the short hash does not work with bitbucket and azure DevOps for example

depends, shorten version is sometimes broken on github (tho i noticed it only 2 times on github enterprise)

dangreen:
Thanks. The current presets now display a short hash while building commit URLs from the full commit hash, so this change is already covered by the current implementation. Closing as completed.

PR #882, opened 2021-12-18
fix: support BREAKING-CHANGE alongside BREAKING CHANGE

jeandeaual:
I noticed a discrepancy between the default configuration of the `conventional-changelog-conventionalcommits` and `conventional-commits-parser` packages and the Conventional Commits specification.

As mentioned in [the Conventional Commits v1.0.0 specification](https://www.conventionalcommits.org/en/v1.0.0/#specification) paragraph 16: “BREAKING-CHANGE MUST be synonymous with BREAKING CHANGE, when used as a token in a footer.”


bcoe:
@jeandeaual this seems like a good fix, but it seems to be breaking some tests.

jeandeaual:
> @jeandeaual this seems like a good fix, but it seems to be breaking some tests.

@bcoe Sorry for that, it should be be fine now.

bcoe (approved):

bcoe:
@jeandeaual thank you for the great contribution :+1:

PR #906, opened 2022-01-19
conventional-changelog-conventionalcommits: defaultConfig use immutable assign

imcuttle:
fix the mutable update operation outer config object

more detail: https://github.com/imcuttle/mometa/commit/c430068d4b91779970591ed33d0131cf9388afdf

image

### DIFF
--- a/packages/conventional-changelog-conventionalcommits/writer-opts.js
+++ b/packages/conventional-changelog-conventionalcommits/writer-opts.js
@@ -175,7 +175,7 @@ function getWriterOpts (config) {
 
 // merge user set configuration with default configuration.
 function defaultConfig (config) {
-  config = config || {}
+  config = Object.assign({}, config)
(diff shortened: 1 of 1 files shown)


packages/conventional-changelog-conventionalcommits/writer-opts.js
```diff
@@ -175,7 +175,7 @@ function getWriterOpts (config) {
 
 // merge user set configuration with default configuration.
 function defaultConfig (config) {
-  config = config || {}
+  config = Object.assign({}, config)
```
armano2, line 178:
```suggestion
  config = { ...config }
```

armano2, line 178: can you use spread here?

```suggestion
function defaultConfig (originalConfig) {
  const config = { ...originalConfig }
```

PR #932, opened 2022-07-03
Remove through2 dep

XaveScor:
Through2 can replaced with native nodes API. I replaced it in this PR
https://nodejs.org/dist./v10.22.0/docs/api/all.html#stream_simplified_construction


dangreen:
@XaveScor Hi. Thank you for the idea. Implemented here: https://github.com/conventional-changelog/conventional-changelog/pull/971

PR #941, opened 2022-09-02
fix(git-semver-tags): escape regexp characters in the tagPrefix option

mattlewis92:
Currently if you're using a `+` symbol in the tagPrefix option it doesn't work as the character isn't escaped when passed to the regexp. This fixes it and adds a test to verify it works correctly.


packages/git-semver-tags/test/index.spec.js
```diff
@@ -257,6 +257,24 @@ describe('git-semver-tags', () => {
...
+    exec('git add --all && git commit -m"eleventh commit"')
+    exec('git tag notms+7.0.0')
+
+    gitSemverTags({ tagPrefix: 'ms+' }, function (err, tags) {
+      if (err) done(err)
+      assert.deepStrictEqual(tags, ['ms+7.0.0', 'ms+6.0.0'])
+      done()
+    })
+  })
+
```
dangreen, line 277: Please see how other tests are written. You should use promises, `testTools` and `expect` util.

packages/git-semver-tags/index.js
```diff
@@ -36,7 +37,7 @@ module.exports = function gitSemverTags (opts, callback) {
     const tags = []
     let tagPrefixRegexp
     if (options.tagPrefix) {
-      tagPrefixRegexp = new RegExp('^' + options.tagPrefix + '(.*)')
+      tagPrefixRegexp = new RegExp('^' + escapeStringRegexp(options.tagPrefix) + '(.*)')
```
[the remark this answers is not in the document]
mattlewis92: Great point, done!

packages/git-semver-tags/test/index.spec.js
mattlewis92: oops, missed the new test format when syncing in the latest master, fixed!

mattlewis92:
@dangreen all changed addressed, thanks for the great feedback!

dangreen (approved):

PR #945, opened 2022-10-01
feat(git-raw-commits): Support multiple path filters

### LINKED ISSUE #944: Support Multiple Path Filters
Some monorepos may have common projects that are shared across the various releasable projects within them.  I would like to be able to use multiple path filters to be able to find all changes that have been made to either a common directory or a specific project's directory.

Git already supports this by passing multiple `-- <path>` arguments.

TheSench:
This PR updates the `gitOpts` object passed to `git-raw-commits` to support multiple values being passed in to the `path` param. This can be done by either passing `gitOpts.path` as an array when using this via JavaScript, or by passing multiple `--path` values via the cli (e.g. `node .\cli.js --path ./CHANGELOG.md --path ./README.md`).

Fixes #944

### DIFF
--- a/packages/git-raw-commits/test.js
+++ b/packages/git-raw-commits/test.js
@@ -227,4 +227,60 @@ describe('git-raw-commits', function () {
           }))
       })
   })
+
+  it('should allow commits to be scoped to a specific set of directories', function (done) {
+    fs.mkdirSync('./packages/common', { recursive: true })
+    writeFileSync('./packages/foo/test1', 'update')
+    exec('git add --all && git commit -m"Fifth commit"')
+    writeFileSync('test2', 'update')
+    exec('git add --all && git commit -m"Sixth commit"')
+    writeFileSync('./packages/common/test1', 'hello')
+    exec('git add --all && git commit -m"Seventh commit"')
+
+    let chunks = []
+    gitRawCommits({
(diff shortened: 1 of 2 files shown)


packages/git-raw-commits/test.js
```diff
@@ -227,4 +227,31 @@ describe('git-raw-commits', function () {
...
+    exec('git add --all && git commit -m"Sixth commit"')
+    writeFileSync('./packages/common/test1', 'hello')
+    exec('git add --all && git commit -m"Seventh commit"')
+
+    let chunks = []
+    gitRawCommits({
+      path: ['./packages/foo', './packages/common'],
+    })
+      .pipe(through(function (chunk, enc, cb) {
+        chunks.push(chunk.toString())
```
TheSench, line 245: 📝 I considered updating the above tests to use this simpler to read format, but felt that would clutter this PR.

I also temporarily updated each test to be able to run in isolation, but that added several seconds to the suite.  I strongly dislike the idea that each test requires the previous one.  Given that the total runtime is still under 10 seconds, that still might be worth doing in a  follow-up PR.

Nazim-hasan:
Why It's not merged yet? Can I contribute?

dangreen:
@Nazim-hasan https://github.com/conventional-changelog/conventional-changelog/pull/1079

PR #974, opened 2023-02-10
refactor!: drop q from dependencies

dangreen:
BREAKING CHANGE: now all promises are native


armano2:
@dangreen #750 you should consider removing "pinkie-promise" as its next package that should be replaced with native promises

PR #977, opened 2023-04-15
fix: simplify retrieval of git-tails in tests

armano2:
Small performance optimization in tests

instead of retrieving all commit hashes we are executing command that is limited to retrival of last element (first in history)


armano2:
fixed by #1067

PR #978, opened 2023-04-16
fix: add possibility to provide custom flags to commit fetcher

### LINKED ISSUE #924: [git-raw-commits] inconsistency in including merge commits
Hi everyone! I faced an interesting thing!

### Issue
I am using tool `@jscutlery/semver` in my nx monorepo for releasing libraries. This tool uses on `conventional-changelog` under the hood.

`@jscutlery/semver` uses `git-raw-commits` to calculate if there is a need in a new version every time there are new commits in master – in my case it happens only on PR merges.
I noticed that **sometimes** merge commit is included into list of commits to calculate changes in a library and **sometimes** not. 

It results in unneeded `patch` releases even when I had only commits of `docs` or `refactor` type(which are ignored by the tool)

### Explanation
`git-raw-commits` uses `git log -- <path>` under the hood.
When path is passed to `git log`, it uses `history simplification` feature which ends up into inconsistent behaviour of `git-raw-commits`.
_Sometimes merge commits are included and sometimes they are not._
[This StackOverflow answer](https://stackoverflow.com/a/56556222/4166537) explains main idea of history simplification

I have created a repo to reproduce this behavior: https://github.com/cakeinpanic/git-raw-commits-demo
It contains 3 simple branches for you to pull, merge locally and observe the tool output

### Proposal
Call `gitRawCommits` with  [--full-history ](https://git-scm.com/docs/git-log#Documentation/git-log.txt---full-history) flag to explicitly include _all the commits_ and avoid inconsistency + add a flag for `conventionalRecommendedBump` to include merge commits into calculations or ignore them completely
```
gitRawCommits({
  format: '%B%H%n',
  from: '1.0.0',
  path: 'folderB',
  'full-history':true
})
```

cakeinpanic:
Providing a way to bypass this issue
closes #924

### DIFF
--- a/packages/conventional-recommended-bump/index.js
+++ b/packages/conventional-recommended-bump/index.js
@@ -16,7 +16,10 @@ function conventionalRecommendedBump (optionsArgument, parserOptsArgument, cbArg
     throw new Error('The \'options\' argument must be an object.')
   }
 
-  const options = Object.assign({ ignoreReverted: true, gitRawCommitsOpts:{} }, optionsArgument)
+  const options = Object.assign({
+    ignoreReverted: true,
+    gitRawCommitsOpts: {}
+  }, optionsArgument)
 
   const cb = typeof parserOptsArgument === 'function' ? parserOptsArgument : cbArgument
 
--- a/packages/conventional-recommended-bump/index.js
+++ b/packages/conventional-recommended-bump/index.js
@@ -16,7 +16,7 @@ function conventionalRecommendedBump (optionsArgument, parserOptsArgument, cbArg
     throw new Error('The \'options\' argument must be an object.')
   }
 
-  const options = Object.assign({ ignoreReverted: true }, optionsArgument)
+  const options = Object.assign({ ignoreReverted: true, gitRawCommitsOpts:{} }, optionsArgument)
 
   const cb = typeof parserOptsArgument === 'function' ? parserOptsArgument : cbArgument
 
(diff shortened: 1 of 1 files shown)


cakeinpanic:
@dangreen please take a look at this small change, it's minor non-breaking release which would unblock me a loooot :)

packages/conventional-recommended-bump/index.js
```diff
@@ -16,7 +16,7 @@ function conventionalRecommendedBump (optionsArgument, parserOptsArgument, cbArg
     throw new Error('The \'options\' argument must be an object.')
   }
 
-  const options = Object.assign({ ignoreReverted: true }, optionsArgument)
+  const options = Object.assign({ ignoreReverted: true, gitRawCommitsOpts:{} }, optionsArgument)
```
dangreen, line 19:
```suggestion
  const options = Object.assign({
    ignoreReverted: true,
    gitRawCommitsOpts: {}
  }, optionsArgument)
```

dangreen (changes requested):
Hi. Thank you for the contribution. 

Could you please run `pnpm fix` to lint and fix possible lint issues?

packages/conventional-recommended-bump/index.js
cakeinpanic: done

cakeinpanic:
@dangreen please take a look now

dangreen:
@bcoe Hello. Could you help with the release? What steps should we take?

dangreen:
@cakeinpanic Released

PR #980, opened 2023-04-24
Update footer.hbs

tomavic:
Is there is any chance to add the `contributors` section. Or allow to show custom section template in the footer?


dangreen:
Thanks for the suggestion. Custom release footers are already supported via `writer.footerPartial` in a JavaScript config; a hard-coded contributors block does not belong in the default preset. This branch is also stale and no longer applies cleanly, so we are closing it.

PR #981, opened 2023-05-05
feat: allow custom note types in changelog

tmeijn:


stevemao:
Can you add a description why this should be `BREAKING CHANGES` instead of `BREAKING CHANGE`?

PR #1018, opened 2023-06-23
fix: avoid overwriting the default values of options with undefined

ryamaguchi0220:
The default values of options (`append`, `releaseCount`, `skipUnstable`...) appear to be not working as intended.

 If `releaseCount` option is not specified, the `options` object initially has the property `releaseCount: undefined`, and then the default value `releaseCount: 1` looks like to be assigned to,  but actually `releaseCount: undefined` is assigned again due to  merging `...options` (which has `releaseCount: undefined`) with the default values.

https://github.com/conventional-changelog/conventional-changelog/blob/11195f2822b389f6b5dc7216c4cf079e0b2883e7/packages/conventional-changelog-core/lib/merge-config.js#L66-L95

This problem has been introduced by #959.


ryamaguchi0220:
@dangreen I would appreciate your review :pray:

dangreen (changes requested):
Thanks for the PR! Please fix CI errors and provide tests for your changes.

ryamaguchi0220:
@dangreen Thanks for your review! 
I added the tests for merge-config [095c28a](https://github.com/conventional-changelog/conventional-changelog/pull/1018/commits/095c28a575902e68d7e254663dcec99de51aee06)

packages/conventional-changelog-core/test/merge-config.spec.js
```diff
@@ -0,0 +1,61 @@
+'use strict'
```
ryamaguchi0220, line 1: the test file names (`merge-config.spec.js` and `index.spec.js`) are based on other packages such as [conventional-commits-parser](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-commits-parser/test), [conventional-changelog-writer](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-writer/test).

packages/conventional-changelog-core/test/merge-config.spec.js
```diff
@@ -0,0 +1,61 @@
...
+    const { options } = await mergeConfig()
+    expect(options).to.include(defaultOptions)
+  })
+
+  it('should return default options when null is passed', async function () {
+    const { options } = await mergeConfig()
+    expect(options).to.include(defaultOptions)
+  })
+
+  it('should return default options when undefined value is passed', async function () {
```
ryamaguchi0220, line 50: If undefined values are not omitted, this test will fail since the default values will be overwritten with undefined.

packages/conventional-changelog-core/lib/merge-config.js
```diff
@@ -53,10 +53,19 @@ function guessNextTag (previousTag, version) {
...
 }
 
+function omitUndefinedValueProps (obj) {
+  if (obj === null || typeof obj !== 'object') {
+    return {}
+  }
+  const omittedObj = Object.entries(obj).filter(([, value]) => value !== undefined)
+  return Object.fromEntries(omittedObj)
+}
+
```
dangreen, line 63: let's optimize it to one loop

```suggestion
function omitUndefinedValueProps (obj) {
  if (!obj) {
    return {}
  }

  const omittedObj = {}
  
  for (const key in obj) {
    if (obj[key] !== undefined) {
      omittedObj[key] = obj[key]
    }
  }
  
  return omittedObj
}

```

dangreen (changes requested):

packages/conventional-changelog-core/lib/merge-config.js
ryamaguchi0220: Optimized in [8f118d5](https://github.com/conventional-changelog/conventional-changelog/pull/1018/commits/8f118d5ee0da625e741012a08f5d8e4f235857f7)

dangreen (approved):

ryamaguchi0220:
@dangreen Thanks for the quick release!

PR #1044, opened 2023-07-09
fix: properly deep clones the commit before modifying it

### LINKED ISSUE #1043: v6 of conventional-changelog-conventionalcommits no longer includes hidden commits with breaking changes
Let me explain.  In v5.0.0, which is currently used in release-please, commits titled:

```
//v5

chore!: update a very important dep
// included in the changelog by virtue of being marked as a breaking change

chore: update dep
// not included.
```

```
//v6

chore!: update a very important dep
// no longer included

chore: update dep
// not included.
```

Is this an intentional change or a bug?

JohnCampionJr:
Includes a test that failed prior to this fix.

fixes #1043

### DIFF
--- a/packages/conventional-changelog-writer/lib/util.js
+++ b/packages/conventional-changelog-writer/lib/util.js
@@ -151,6 +151,30 @@ function immutableSet (context, path, value) {
   }
 }
 
+function cloneCommit (commit) {
+  if (!commit || typeof commit !== 'object') {
+    return commit
+  } else
+  if (Array.isArray(commit)) {
+    return commit.map(cloneCommit)
+  }
+
+  const commitClone = {}
+  let value
+
+  for (const key in commit) {
+    value = commit[key]
+
+    if (typeof value === 'object') {
+      commitClone[key] = cloneCommit(value)
+    } else {
+      commitClone[key] = value
+    }
+  }
+
+  return commitClone
+}
+
 function processCommit (chunk, transform, context) {
   let commit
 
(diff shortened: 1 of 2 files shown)


packages/conventional-changelog-writer/lib/util.js
```diff
@@ -158,9 +158,8 @@ function processCommit (chunk, transform, context) {
     chunk = JSON.parse(chunk)
   } catch (e) {}
 
-  commit = {
-    ...chunk
-  }
+  // ensure two 100% separate objects (for saving to raw)
+  commit = JSON.parse(JSON.stringify(chunk))
```
dangreen, line 162: This way is very very slow. Let's write function to clone commit object:

```js
function cloneCommit(commit) {
  if (!commit || typeof commit !== 'object') {
    return commit
  } else
  if (Array.isArray(commit)) {
    return commit.map(cloneCommit)
  }

  const commitClone = {}
  let value

  for (const key in commit) {]
    value = commit[key]

    if (typeof value === 'object' ) {
      commitClone[key] = cloneCommit(value)
    } else {
      commitClone[key] = value
    }
  }

  return commitClone
}
```

packages/conventional-changelog-writer/lib/util.js
```diff
@@ -158,9 +158,8 @@ function processCommit (chunk, transform, context) {
...
 
   if (typeof transform === 'function') {
     commit = transform(commit, context)
```
dangreen, line 165:
```suggestion
    // @todo Make transformers immutable
    commit = transform(commit, context)
```

dangreen (changes requested):
Thanks for the PR!

JohnCampionJr:
Changes incorporated!

dangreen (approved):

PR #1063, opened 2023-07-15
feat(raw-commits): allow hidding by subject pattern

roggervalf:
Add a way to hide commits with specific pattern from the release notes. This is useful in order to avoid commits with specific messages that we don't want to include in release notes


roggervalf:
please @dangreen when you get a chance 👀

dangreen:
@roggervalf Hi. I think "types" is not right place for filtering by subject. We can add new option "ignore" to git-raw-commits to ignore commits by regex ([here](https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/git-raw-commits/index.js#L68)). And pass it option from preset

https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-conventionalcommits/index.js#L15

```
{
  gitRawCommitsOpts: {
    ignore: config.ignoreCommits,
    noMerges: null
  }
}
```

WDYT?

roggervalf:
> @roggervalf Hi. I think "types" is not right place for filtering by subject. We can add new option "ignore" to git-raw-commits to ignore commits by regex ([here](https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/git-raw-commits/index.js#L68)). And pass it option from preset
> 
> https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-conventionalcommits/index.js#L15
> 
> ```
> {
>   gitRawCommitsOpts: {
>     ignore: config.ignoreCommits,
>     noMerges: null
>   }
> }
> ```
> 
> WDYT?

Thank you @dangreen, sounds good to me, I think I could address it

packages/git-raw-commits/index.js
```diff
@@ -65,7 +65,9 @@ function gitRawCommits (rawGitOpts, rawExecOpts) {
         transform (chunk, enc, cb) {
           isError = false
           setImmediate(() => {
-            readable.push(chunk)
+            if ( !(gitOpts?.ignore && isMatch(chunk.toString(), gitOpts?.ignore))) {
```
dangreen, line 68: Let's use regex instead of glob

```js
const ignoreRegex = typeof gitOpts.ignore === 'string'
  ? new RegExp(gitOpts.ignore)
  : gitOpts.ignore
const shouldNotIgnore = ignoreRegex
  ? chunk => !ignoreRegex.test(chunk.toString())
  : () => true
...
if (shouldNotIgnore(chunk)) {
  readable.push(chunk)
}
```

dangreen (changes requested):
Ignore commit by pattern is a rare case, so I don't want to add new dependency for it. Regex should fit you needs

roggervalf:
Thank you @dangreen, it sound good to me

dangreen (approved):

roggervalf:
hey @dangreen one question, will this change be promoted on version 7 or could it be promoted on a minor version (6.x)?

PR #1064, opened 2023-07-15
fix(types): allow combining provided types with defaults

roggervalf:


roggervalf:
please @dangreen when you get a chance :eyes:

roggervalf:
please @dangreen when you get a chance :eyes:

dangreen:
@roggervalf I'll take a look this week 👌

dangreen:
@roggervalf Hi. Sorry for the delay. I want to suggest more easier way: what if we will export default types, so you can extend / modify them 

```js
const preset = require('conventional-changelog-conventionalcommits')

conventionalChangelogCore({
  config: preset({
    types: preset.defaultCommitTypes.map((commitType) => (
      commitType.type === 'chore'
        ? { ...commitType, hidden: false }
        : commitType
    ))
  })
})
```

WDYT?

roggervalf:
> @roggervalf Hi. Sorry for the delay. I want to suggest more easier way: what if we will export default types, so you can extend / modify them
> 
> ```js
> const preset = require('conventional-changelog-conventionalcommits')
> 
> conventionalChangelogCore({
>   config: preset({
>     types: preset.defaultCommitTypes.map((commitType) => (
>       commitType.type === 'chore'
>         ? { ...commitType, hidden: false }
>         : commitType
>     ))
>   })
> })
> ```
> 
> WDYT?

hey @dangreen, but in that case how it will work with commit-analyzer usage https://github.com/semantic-release/commit-analyzer#usage? the configuration is passed in a json file

dangreen:
@roggervalf Configuration can be made as js file: https://github.com/semantic-release/semantic-release/blob/master/docs/usage/configuration.md#configuration

> A .releaserc file, written in YAML or JSON, with optional extensions: .yaml/.yml/.json/.js/.cjs
> A release.config.(js|cjs) file that exports an object

roggervalf:
Thank you @dangreen, it sounds good to me

packages/conventional-changelog-conventionalcommits/test/index.spec.js
```diff
@@ -193,6 +193,33 @@ describe('conventional-changelog-conventionalcommits', () => {
...
+        expect(chunk).to.include('### Miscellaneous Chores')
+        expect(chunk).to.include('**deps:** upgrade example from 1 to 2')
+
+        expect(chunk).to.include('### Performance Improvements')
+        expect(chunk).to.include('**ngOptions:** make it faster')
+
+        done()
+      }))
+  })
+
```
dangreen, line 222: Please use arrow functions and `for await`. You can find examples in this test file.

dangreen (changes requested):

packages/conventional-changelog-conventionalcommits/test/index.spec.js
roggervalf: It should be addressed, thank you

roggervalf: It should be addressed, thank you

dangreen (approved):

PR #1070, opened 2023-08-02
ci: use node 16 and pnpm 8

dangreen:


dangreen:
@bcoe Hi. Could you please remove "test (14)" from required jobs and make "test (20)" required?

PR #1074, opened 2023-08-06
feat: migrate from chalk to picocolors

dangreen:


escapedcat:
You're doing a great job! Should do this with commilint as well. Is picocolors a drop-in replacement for chalk?

dangreen:
@escapedcat Hi. Almost, it has different api but features are the same.

escapedcat:
Ugh, need to check. Thanks!
Impressed by your motivation really.

PR #1090, opened 2023-08-26
refactor(conventional-commits-parser): use Transform instead of JSONStream

armano2:
follow up for #971

Use simple stream transform instead of JSONStream library,

- [JSONStream](https://github.com/dominictarr/JSONStream) is archived and no longer maintained.
- [through](https://github.com/dominictarr/through) is no longer included in dependency tree


dangreen (approved):

armano2:
tip: you can use squash instead of rebasing/force pushing changes manually

PR #1113, opened 2023-08-31
refactor: rewrite split2 to simple transform and inline code

armano2:


packages/git-raw-commits/index.js
```diff
@@ -62,13 +61,15 @@ async function streamRawCommits (readable, gitOpts, execOpts) {
   })
 
   ;(async () => {
-    for await (const chunk of child.stdout.pipe(split(DELIMITER + '\n'))) {
+    for await (const chunk of child.stdout) {
       isError = false
 
-      await immediate()
+      for await (const line of chunk.toString().split(DELIMITER + '\n')) {
```
armano2, line 67: there must be a better way to do this, 

```ts
function splitStream (match) {
  return new Transform({
    readableObjectMode: true,
    transform (chunk, enc, cb) {
      for (const line of chunk.toString().split(match)) {
        this.push(line)
      }
      cb()
    }
  })
}
```

PR #1115, opened 2023-09-01
feat(conventional-changelog-preset-loader): throw error if preset does not export a function

dangreen:


1aron:
@dangreen If the preset format from the old to the new remains consistent, the only difference being the additional support for the factory function, would it be preferable to maintain compatibility with the old version?

This change would require third-party conventional-changelog-* presets to align with the new version and release updates, which may result in a loss of compatibility for existing presets in the current market.

PR #1120, opened 2023-09-05
fix(conventional-changelog-conventionalcommits): optional config

lvjiaxuan:
It throws error, when the preset option is set to the string `conventionalcommits`.

image

### DIFF
--- a/packages/conventional-changelog-conventionalcommits/conventionalRecommendedBump.js
+++ b/packages/conventional-changelog-conventionalcommits/conventionalRecommendedBump.js
@@ -27,7 +27,7 @@ function createConventionalRecommendedBumpOpts (config, parserOpts) {
         }
       })
 
-      if (config.preMajor && level < 2) {
+      if (config?.preMajor && level < 2) {
         level++
       }
 
--- a/packages/conventional-recommended-bump/test/index.spec.js
+++ b/packages/conventional-recommended-bump/test/index.spec.js
@@ -71,13 +71,24 @@ describe('conventional-recommended-bump', () => {
     await expect(() => conventionalRecommendedBump({ cwd: testTools.cwd }, {})).rejects.toThrow()
   })
 
+  it('should allow the string `conventionalcommits` as the preset option', async () => {
+    preparing(2)
+
+    await expect(() => conventionalRecommendedBump({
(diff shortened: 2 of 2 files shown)


packages/conventional-changelog-conventionalcommits/conventionalRecommendedBump.js
```diff
@@ -27,7 +27,7 @@ function createConventionalRecommendedBumpOpts (config, parserOpts) {
         }
       })
 
-      if (config.preMajor && level < 2) {
+      if (config?.preMajor && level < 2) {
```
dangreen, line 30: Please add test case for that.

packages/conventional-recommended-bump/test/index.spec.js
```diff
@@ -77,9 +77,7 @@ describe('conventional-recommended-bump', () => {
 
       const recommendation = await conventionalRecommendedBump({
         cwd: testTools.cwd,
-        preset: {
-          name: 'conventionalcommits'
-        }
+        preset: 'conventionalcommits'
       }, {})
```
dangreen, line 81: Why did you change it?

dangreen (changes requested):
Hi. Thanks for the PR!

packages/conventional-recommended-bump/test/index.spec.js
lvjiaxuan: Because this string preset causes the issue and performs the same thing as `{  name: 'conventionalcommits'  }`. If you think this is inappropriate, I will write an additional test case for it.

packages/conventional-recommended-bump/test/index.spec.js
```diff
@@ -71,6 +71,15 @@ describe('conventional-recommended-bump', () => {
...
 
+  it('should allow the string `conventionalcommits` as the preset option', async () => {
+    preparing(2)
+
+    await expect(() => conventionalRecommendedBump({
+      cwd: testTools.cwd,
+      preset: 'conventionalcommits'
+    }, {})).not.toThrowError()
+  })
+
```
lvjiaxuan, line 82: Finally, I decided to add a test case. 😁

dangreen (approved):

PR #1126, opened 2023-09-08
feat(conventional-commits-parser)!: rewrite to TypeScript

dangreen:
BREAKING CHANGES: Now package exports CommitParser class and parseCommitsStream method for backward compatability.

PR #1150, opened 2023-09-11
feat(conventional-changelog-writer)!: rewrite to TypeScript

dangreen:


packages/conventional-changelog-writer/index.js
```diff
@@ -63,185 +66,166 @@ async function conventionalChangelogWriterInit (context, options) {
     ...options
   }
 
-  if (!options.transform || typeof options.transform === 'object') {
-    options.transform = {
-      hash: function (hash) {
+  if (!finalOptions.transform || typeof finalOptions.transform === 'object') {
+    finalOptions.transform = {
+      hash: (hash) => {
```
armano2, line 71:
```suggestion
      hash (hash) {
```

dangreen: https://github.com/conventional-changelog/conventional-changelog/pull/1126/files#diff-6884918dc8291219be508e05e28965b958c734def85324f3b53858ea4702090fR2

https://eslint.org/docs/latest/use/configure/configuration-files#using-configuration-files

there is `"root": true` property in root eslint config, so all others configs are extending parent config

Issue #1153, opened 2023-10-11
conventional-changelog-cli: context file type must json, not support js file


damingerdai:
i found in this pr: [fix: fix semver vulnerability] (https://github.com/conventional-changelog/conventional-changelog/pull/1071) has a big break change that we can not use javescripts file as context file. we must use json file to setup context info.

so my question is that cloud your revert it these change? If no, why and please update it changelog file?

```diff
-  templateContext = JSON.parse(await readFile(relativeResolve(flags.context), 'utf8'))
+ templateContext = require(resolve(process.cwd(), flags.context))
```

stevemao:
Feel free to create a PR to support both

PR #1156, opened 2023-10-15
revert(conventional-changelog-cli): contenxt supports both json and js file types at the same time

damingerdai:
close: #1153


dangreen (changes requested):
@damingerdai Hi. ESM modules doesn't support `require` method. This code will not work

PR #1159, opened 2023-10-24
fix(conventional-changelog-cli): wrong link for option prompt

barnett617:
This is clearly breaked due to https://github.com/conventional-changelog/conventional-changelog/pull/1144 which missed this link's update


dangreen (approved):

dangreen:
Thanks!

PR #1167, opened 2023-11-18
feat(conventional-commits-parser): add bodyLines to commit object

### LINKED ISSUE #1166: Add array representation of commit.body to conventional-commits-parser
I'd like to propose adding a new property `bodyLines` to the `commit` object created by conventional-commits-parser. It would be a line-by-line representation of `commit.body`.

I'd be happy to try and contribute this if there is an interest in having it. I created a draft of this feature and it seemed relatively straightforward to implement.

## Motivation

I was trying to customize the changelog generated in my project during a semantic-release run. I'm doing so by providing `writerOpts.commitPartial` for the `conventionalcommits` preset for `@semantic-release/release-notes-generator`. This works well in principle.

Specifically, I want to add the commit body to the end of [the commit.hbs template](https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-writer/templates/commit.hbs) by appending something like this:

```hbs
{{#if body}}

  {{body}}

{{/if}}
```

This works; however, due to the nature of `commit.body` being a single string, I can't control the indentation if the body consists of multiple lines and so I'm getting a changelog entry like this:

```plaintext
    * feat: support all angle values as input

      Adds support for the angle value units deg, grad, rad, and turn when entering hues (see https://www.w3.org/TR/css-values-4/#angle-value).

Stops normalizing angle values to the range [0, 360) (e.g. a hue value of 450 will no longer be processed, stored, and emitted as 90).
```

With the addition I'm proposing, I would instead use the following:

```hbs
{{#each bodyLines}}

  {{!-- What would be the right syntax here if `bodyLines` is of type `string[]`? --}}
  {{this}}

{{/each}}
```

and get the following entry (note the difference in indentation of the last line):

```plaintext
    * feat: support all angle values as input

      Adds support for the angle value units deg, grad, rad, and turn when entering hues (see https://www.w3.org/TR/css-values-4/#angle-value).

      Stops normalizing angle values to the range [0, 360) (e.g. a hue value of 450 will no longer be processed, stored, and emitted as 90).
```

kleinfreund:
Adds the new property `bodyLines` as a line-by-line representation of `body` to the commit object.

Closes #1166.

### DIFF
--- a/packages/conventional-changelog-writer/README.md
+++ b/packages/conventional-changelog-writer/README.md
@@ -27,6 +27,7 @@ It expects an object mode upstream that looks something like this:
   scope: 'scope',
   subject: 'broadcast $destroy event on scope destruction',
   body: null,
+  bodyLines: [],
   footer: 'Closes #1',
   notes: [],
   references: [ { action: 'Closes', owner: null, repository: null, issue: '1', raw: '#1' } ] }
@@ -36,6 +37,7 @@ It expects an object mode upstream that looks something like this:
   scope: 'ng-list',
   subject: 'Allow custom separator',
   body: 'bla bla bla',
+  bodyLines: ['bla bla bla'],
   footer: 'BREAKING CHANGE: some breaking change',
   notes: [ { title: 'BREAKING CHANGE', text: 'some breaking change' } ],
   references: [] }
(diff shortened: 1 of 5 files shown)


dangreen:
https://github.com/conventional-changelog/conventional-changelog/issues/1166#issuecomment-1817589185

PR #1169, opened 2023-11-19
fix(angular): revert pattern to not require a period

privatenumber:
As per their spec, there is no period required at the end of the commit SHA:
https://github.com/angular/angular/blob/main/CONTRIBUTING.md#revert-commits

Also, as a fix, I set a minimum requirement of 7 characters for the SHA ([Git default](https://stackoverflow.com/a/18134919)) as currently you can pass in nothing and it would still work.

### DIFF
--- a/packages/conventional-changelog-angular/src/parser.js
+++ b/packages/conventional-changelog-angular/src/parser.js
@@ -7,7 +7,7 @@ export function createParserOpts () {
       'subject'
     ],
     noteKeywords: ['BREAKING CHANGE'],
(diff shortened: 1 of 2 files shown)


dangreen (changes requested):
Hi. Thank you for the PR!

Please fix tests. 

> you can pass in nothing and it would still work.

Also please add test case for that

privatenumber:
Couldn't figure out how to make the negative test work for SHAs less than 7 characters or more than 40 characters. Something was still picking up the `revert:` prefix without the "This reverts commit" body.

Would you mind looking into it?

PR #1189, opened 2023-12-22
feat(conventional-changelog-writer)!: `formatDate` option

### LINKED ISSUE #1186: Misuse of Swedish Locale (conventional-changelog-writer/src/utils.ts)
Code at https://github.com/conventional-changelog/conventional-changelog/blob/7b13ec9afcbe7afc24ce47af805c339ba7f0276f/packages/conventional-changelog-writer/src/utils.ts#L11-L25 (#1162)  misuses the Swedish locale for non-internationalization purposes. Don't do that if you expect identical behavior, locale data can and does change.

Better would be to take the user's locale and use it, if the output is for humans, or format some other way.

Referenced from https://github.com/nodejs/node/issues/51090

dangreen:
BREAKING CHANGES: `timeZone` option was removed, please use `formatDate` option to change timezone of a date.

closes #1186

### DIFF
--- a/packages/conventional-changelog-writer/src/utils.ts
+++ b/packages/conventional-changelog-writer/src/utils.ts
@@ -3,25 +3,17 @@ import type {
   StringsRecord
 } from './types/index.js'
 
-const utcDateFormatter = Intl.DateTimeFormat('sv-SE', {
-  timeZone: 'UTC'
-})
+const DATETIME_LENGTH = 10
 
 /**
  * Formats date to yyyy-mm-dd format.
- * @param date - Date string, number or Date object.
- * @param timeZone - Time zone to use.
+ * @param date - Date string or Date object.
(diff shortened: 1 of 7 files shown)


packages/conventional-changelog-writer/src/utils.ts
```diff
@@ -3,25 +3,17 @@ import type {
...
  */
-export function formatDate(date?: string | number | Date, timeZone = 'UTC') {
-  const dateFormatter = !timeZone || timeZone === 'UTC'
-    ? utcDateFormatter
-    : Intl.DateTimeFormat('sv-SE', {
-      timeZone
-    })
-
-  // sv-SEis used for yyyy-mm-dd format
-  return dateFormatter.format(date ? new Date(date) : new Date())
```
srl295, line 24: Good riddance!

PR #1215, opened 2024-03-10
feat!: cleanup presets interface

dangreen:
BREAKING CHANGES: now preset object provides `commits`, `parser`, `writer` and `whatBump` properties.


knocte:
@dangreen I'm just curious, when do you plan to release a new version with this breaking change?

dangreen:
@knocte We have [milestone for next release](https://github.com/conventional-changelog/conventional-changelog/milestone/4), but I can't name a date, because it depends on my free time.

knocte:
> but I can't name a date, because it depends on my free time.

I'm of course not asking for it to be earlier rather than later. If anything, it would be the other way around; because I'm trying to understand why the last version of this pkg was released so long ago: https://www.npmjs.com/package/conventional-commits-parser?activeTab=versions

dangreen:
@knocte Because there are packages that depends on conventional-commits-parser that also should be updated. So I have plan to publish all updated packages at once.

PR #1218, opened 2024-03-16
feat(conventional-recommended-bump)!: new `Bumper` exported class

dangreen:
BREAKING CHANGES: old exported function was removed, please read docs how to use new `Bumper` exported class.


ghiscoding:
@dangreen it's nice to see so many conventional-changelog changes recently but it's hard to follow especially since there's no migration info like this package. I'm trying to apply the changes to Lerna-Lite (a lighter version of Lerna) but from I see, it looks like all the options used by Lerna like `lernaPackage`, `path`, `tagPrefix` are now only accessible from the CLI? That's not possible in the case of Lerna/Lerna-Lite since it must be executed dynamically by code and preset can also be dynamic in Lerna so CLI is not an option. All that to say that I'm confused with the migration even after taking a look at the content of this PR

This is part of what Lerna's code looks like

```ts
  const options: conventionalRecommendedBump.Options = {
    path: pkg.location,
  };

  if (type === 'independent') {
    options.lernaPackage = pkg.name;
  } else {
    // only fixed mode can have a custom tag prefix
    options.tagPrefix = tagPrefix;
  }

const data = await conventionalRecommendedBump(options);
let releaseType = data.releaseType || 'patch';
const newVer = semver.inc(pkg.version, prereleaseType, prereleaseId), buildMetadata)
```
all I seem to get is this piece of code and I don't understand where to put the options
```ts
const bumper = new Bumper(pkg.location);
const recommendation = await bumper.bump();
```

dangreen:
@ghiscoding https://github.com/conventional-changelog/conventional-changelog/issues/1266#issuecomment-2105633517

PR #1226, opened 2024-04-18
feat!: drop node 16 support

dangreen:
BREAKING CHANGE: Node >= 18 is required


dangreen:
@knocte Some of our dependencies have droped node 16 support

PR #1233, opened 2024-04-22
feat(conventional-commits-parser): add flag to enable/disable multiple-scopes (default: enabled)

wlc015f:
This PR close to [issue 4015](https://github.com/conventional-changelog/commitlint/issues/4015)

migrate separating scopes logic code from `@commitlint/rules/src/scope-enum.ts`.

to Implement the feature add `isEnableMultipleScopes` and `scopeDelimitersPattern` parserOptions:

 - `isEnableMultipleScopes` used to determine whether to separate scopes by slash or comma delimiters contain in scope.

 - `scopeDelimitersPattern` used to separate scopes if `isEnableMultipleScopes` value is true.

the two default parserOptions values of two parserOptions refer to [@commitlint/rules/src/scope-enum.ts](https://github.com/conventional-changelog/commitlint/blob/aa4df36b1d07e56230ff569aec5af727f1144aae/%40commitlint/rules/src/scope-enum.ts#L16), this pr make it configurable, and work as before if not config

### DIFF
--- a/packages/conventional-commits-parser/src/CommitParser.spec.ts
+++ b/packages/conventional-commits-parser/src/CommitParser.spec.ts
@@ -90,6 +90,37 @@ describe('conventional-commits-parser', () => {
         expect(result.scope).toBe('hello/world')
         expect(result.subject).toBe('message')
       })
+
+      it('should parse scopes with delimiters when default isEnableMultipleScopes and scopeDelimitersPattern option', () => {
+        const parser = new CommitParser({
+          headerPattern: /^(\w*)(?:\((.*)\))?: (.*)$/ // TODO default headerPattern do not match scopeDelimiters
+        })
+        const commitsWithScopeDelimiters = [
+          {
+            commit: 'feat(hello/world): message',
+            parsedScore: 'hello/world'
+          },
+          {
+            commit: 'feat(hello\\world): message',
+            parsedScore: 'hello\\world'
+          },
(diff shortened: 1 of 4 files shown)


knocte:
Why do we need a flag at all? Why not just assume multiple scopes? Do we have a unit test that would fail if this is done without a flag?

wlc015f:
> Why do we need a flag at all? Why not just assume multiple scopes? Do we have a unit test that would fail if this is done without a flag?

our projects previous use scope like: `feat(A/B): commit msg`, B is a sub scope of A, when we update `@commitlint/cli` from 8.3.5 to 19.2.1, it doesn't work well as before, because for version 19.2.1 "/" is scopes delimiter to separate input scope string into several scopes.  `scopeEnum` function in  `@commitlint/rules` would separate input scope "A/B" to scope "A" and "B", scope "A/B" can not pass scopeEnum check.

so i want to migrate separating scopes logic code from `@commitlint/rules/src/scope-enum.ts` to  package`conventional-commits-parser`, set default behavior work as before in `@commitlint/rules`, and then set `isEnableMultipleScopes`=false can solve my problem.

knocte:
Why was this closed?

wlc015f:
> Why was this closed?

this feature is a fix for a very old version. Waiting for the next+2 roadmap update there might bring better planning and implementation.

PR #1277, opened 2024-05-20
fix(git-client): match semver better in unstable tag regex

### LINKED ISSUE #935: unstableTagTest doesn't support empty preid
# WHAT'S THE PROBLEM
unstableTagTest regex doesn't support pre-releases with empty `preid`

## examples
✅  1.0.2-re.2 
❌  1.0.2-1

### LINKED ISSUE #1276: unstableTagRegex isn't semver compliant
https://github.com/conventional-changelog/conventional-changelog/blob/d3b8aaa16337993bbad4d91db1ffac5a7568756b/packages/git-client/src/ConventionalGitClient.ts#L94

This regex doesn't take into account that semver defines pre-release like this "A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers immediately following the patch version..." https://semver.org/#spec-item-9

Better regex could be like `[1-9]\d*\.[1-9]\d*\.[1-9]\d*-` so it still doesn't care about prefix but only about the hyphen after 3 numbers separated by dot.

My problem is that someone used semver compatible versioning with pre-releases as such: 
```
v1.1.1-rc1-13
v1.1.1-rc1-14
v1.1.1-rc1-15
```
and the current regex doesn't catch this.

Rulqu:
Fixes #1276
Fixes #935


dangreen:
@Rulqu Thank you! please add test case for that changes

Rulqu:
@dangreen I don't have any experience in how to do tests. I'm just a devops who wanted to use this tool and knew a bit of regex. I could learn but I don't really know how to start.

dangreen:
@Rulqu Looks like your regex is incorrect. Please see tests result

Rulqu:
@dangreen Ah, I figured out how the testing works here. I guess I didn't research enough. I fixed the regex and added some entries to the test that failed before my fix.

Rulqu:
This also fixes #935 and invalidates #936

dangreen (approved):

PR #1281, opened 2024-05-31
fix(git-client): close the generator used by getLastSemverTag

Lukinoh:
Hello,

I encountered a bug with `conventional-recommended-bump@10.0.0`.
I am using it programmatically.

Concretely, after running my code that generates the bump, the node program does not exit.
You have to `CTRL^C`.

I was not able to reproduce it with a small repository or with a test, but still I found a way to reproduce it with another repository than mine.

<details>
  <summary>To reproduce it, you have to follow these steps (click me)</summary>
  
  1. `git clone git@github.com:vitejs/vite.git`
  1. `cd vite`
  1. `mkdir bump`
  1. `cd bump`
  1. `touch package.json`
  1. Add the following text to `package.json`
  
  ```json
  {
    "type": "module",
    "dependencies": {
      "conventional-recommended-bump": "10.0.0",
      "conventional-changelog-angular": "8.0.0"
    }
  }
  
  ```
  7. `touch index.js`
  8. Add the following code to `index.js`
  ```javascript
  import { Bumper } from 'conventional-recommended-bump';
  
  const bumper = new Bumper()
  bumper.loadPreset('angular')
  bumper.commits({
    path: ['../']
  })
  bumper.tag({
    prefix: "plugin-legacy@"
  })
  
  const bump = await bumper.bump()
  
  console.log(bump)
  console.log("blocked");
  ```
  9. `npm install`
  10. `node ./index.js`
</details>

---

From my investigation, the culprit code is this:
https://github.com/conventional-changelog/conventional-changelog/blob/d3b8aaa16337993bbad4d91db1ffac5a7568756b/packages/git-client/src/ConventionalGitClient.ts#L137-L139

I suppose that in some (big?) repositories, as the generator never finished, it was preventing the node process to finish.

As I mentioned before, I was not able to create a test for it as I was not able to reproduce it using tests.
But the solution fixes the issue in my repository and vite's one.

If you have any other idea, let me know.


packages/git-client/src/ConventionalGitClient.ts
```diff
@@ -135,7 +135,11 @@ export class ConventionalGitClient extends GitClient {
    * @returns Last semver tag, `null` if not found.
    */
   async getLastSemverTag(params: GetSemverTagsParams & Params = {}) {
-    return (await this.getSemverTags(params).next()).value || null
+    const semverTagsStream = this.getSemverTags(params)
+    const lastSemverTag = (await semverTagsStream.next()).value || null
+
+    await semverTagsStream.return()
+    return lastSemverTag
```
dangreen, line 142:
```suggestion
    const stream = this.getSemverTags(params)
   
    for await (const tag of stream) {
      return tag
    }
   
    return null
```

I think it more readable

dangreen (changes requested):
Thank you for the PR!

Please also apply this fix to GitClient#getLastTag

packages/git-client/src/ConventionalGitClient.ts
Lukinoh: I agree with you.
I also was not happy with the readability of the previous solution.

I extracted the code in a function to avoid duplication and also have only one `// eslint-disable-next-line no-unreachable-loop`.
I've added tests for it.

I added a test for `getLastTag` and `getLastSemverTags`, not sure they are very useful.
Let me know if you want to keep them or not.

PR #1285, opened 2024-06-07
feat: prevent modifications to Date objects in commit.ts

MHDFahz:
This change introduces a condition within the `preventModifications` function in the `commit.ts` file. The purpose of this condition is to check if the object being processed is an instance of `Date`. 

In JavaScript, when a `Date` object is converted to a string (for instance, when using template literals or calling `toISOString`), the language internally tries to access and modify some properties of the `Date` object. If the `Date` object has been made immutable by a function like `preventModifications`, these operations result in a `TypeError`.

The specific error thrown is: "semantic-release: TypeError: Method Date.prototype.toString called on incompatible receiver [object Date]". This error was originally reported by `semantic-release`.

To circumvent this error, the introduced condition checks if the object is an instance of `Date`. If it is, the function immediately returns the object as is, thereby avoiding the error.

This change enhances the stability and reliability of the code by preventing a known error scenario. It specifically addresses the TypeError that occurs when JavaScript attempts to convert an immutable `Date` object to a string.


mlazari:
@MHDFahz @dangreen Was this ever released? It looks like there was an attempt to publish a new version 8.0.1 for `conventional-changelog-writer` but it failed? I think this would probably fix https://github.com/semantic-release/semantic-release/issues/3571

mlazari:
I just did a [patch](https://www.npmjs.com/package/patch-package) with the change in this PR and it fixed the error for me:

[conventional-changelog-writer+8.0.0.patch](https://github.com/user-attachments/files/18602951/conventional-changelog-writer%2B8.0.0.patch)

dangreen:
@mlazari https://github.com/conventional-changelog/conventional-changelog/issues/1338

PR #1289, opened 2024-07-05
fix(conventional-changelog-core): replace `read-package-up` with lighter alternative

SuperchupuDev:
Replaces `read-package-up` and `read-pkg` with `pkg-types`, bringing down core's dependency count from [45](https://npmgraph.js.org/?q=conventional-changelog-core@8.0.0) to 31 dependencies.

`read-pkg` only really parsed json by calling `normalize-package-data`, which is already used in the project, so no reason to keep `read-pkg` as it's [really](https://npmgraph.js.org/?q=read-pkg%409.0.1) bloated, (and it's also used by `read-package-up`).

Instead, this PR calls `normalize-package-data` directly, and uses `pkg-types` for finding the closest package.json file, which only has [5 dependencies](https://npmgraph.js.org/?q=pkg-types) instead of `read-package-up`'s [28](https://npmgraph.js.org/?q=read-package-up%4011.0.0)


dangreen (approved):

SuperchupuDev:
fixed lockfile conflicts :pray:

PR #1317, opened 2024-12-11
Update README.md

prasanthlouis:
I think this should be a colon not a comma? Possibly?

It should be "owner":null not "owner",null


dangreen (approved):

dangreen:
Thanks!

PR #1342, opened 2025-02-06
chore: write git-raw-commits in TypeScript

okuryu:
Hi, I have written an implementation of the git-raw-commits package in TypeScript. Can you review it?


dangreen:
@okuryu Hi. git-raw-commits is a legacy package. If you need Node.js API with types then use [@conventional-changelog/git-client](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/git-client)

Issue #1371, opened 2025-05-03
headerPattern ignores named capture groups


misogihagi:
The current commitlintparser only accepts positional matches.
However, some people use Named capturing group.
I would like to have Named capturing group too.

https://github.com/arvinxx/gitmoji-commit-workflow/blob/master/packages/parser-opts/src/index.ts#L10
https://github.com/ccnnde/commitlint-config-git-commit-emoji

Because the "?" causes positional problems that the next "?" element is undefined.

example:

```javascript
 const parser = new CommitParser({
  headerPattern: /^:(?<type>.+):(\((?<scope>.+)\))?:\s*(?<subject>.+)$/,
  headerCorrespondence: [
    'type',
    'scope',
    ''subject'
   ]
})

const commit = parser.parse(':type::subject')
commit.subject // undefined!
```

PR #1372, opened 2025-05-03
feat(conventional-commits-parser): named capturing groups support for regex patterns

misogihagi:
#1371

### DIFF
--- a/packages/conventional-commits-parser/src/CommitParser.spec.ts
+++ b/packages/conventional-commits-parser/src/CommitParser.spec.ts
@@ -277,23 +277,10 @@ describe('conventional-commits-parser', () => {
           ]
         })
 
-        function positiveUtil(commit, tokens) {
+        function negativeUtil<T>(commit: T, tokens: (keyof T)[]) {
           tokens.forEach((token) => {
-            expect(commit[token]).toEqual(token)
+            expect(commit[token]).toBeUndefined()
           })
-        }
-
-        function negativeUtil(commit, tokens) {
-          tokens.forEach((token) => {
-            expect(commit[token]).toBeUndefined(token)
-          })
-        }
-
-        function util(commit) {
-          return {
-            positiveUtil: tokens => positiveUtil(commit, tokens),
-            negativeUtil: tokens => negativeUtil(commit, tokens)
-          }
         }
 
         describe('positive test', () => {
@@ -301,26 +288,28 @@ describe('conventional-commits-parser', () => {
           it('should be parsed for normal use cases', () => {
             const commit = parser.parse(':type:(scope):subject')
 
-            util(commit).positiveUtil([
-              'type',
-              'scope',
-              'subject'
-            ])
+            expect(commit).toMatchObject({
+              type: 'type',
+              scope: 'scope',
+              subject: 'subject'
+            })
           })
 
           it('should be parsed even missing scope', () => {
             const commit = parser.parse(':type::subject')
 
-            util(commit).positiveUtil(['type', 'subject'])
+            expect(commit).toMatchObject({
+              type: 'type',
+              subject: 'subject'
+            })
           })
         })
 
         describe('negative test', () => {
-
           it('should not be parsed if the first colon is missing', () => {
             const commit = parser.parse('type:(scope):subject')
 
-            util(commit).negativeUtil([
+            negativeUtil(commit, [
               'type',
               'scope',
               'subject'
@@ -330,7 +319,7 @@ describe('conventional-commits-parser', () => {
           it('should not be parsed if a colon is missing after scope', () => {
             const commit = parser.parse(':type:(scope)subject')
 
-            util(commit).negativeUtil([
+            negativeUtil(commit, [
               'type',
               'scope',
               'subject'
@@ -340,7 +329,7 @@ describe('conventional-commits-parser', () => {
           it('should not be parsed if parentheses are missing before or after a scope.', () => {
             const commit = parser.parse(':type:scope: subject')
 
-            util(commit).negativeUtil([
+            negativeUtil(commit, [
               'type',
               'scope',
               'subject'
@@ -350,7 +339,7 @@ describe('conventional-commits-parser', () => {
           it('should not be parsed if the parentheses are empty', () => {
             const commit = parser.parse(':type:(): subject')
 
-            util(commit).negativeUtil([
+            negativeUtil(commit, [
               'type',
               'scope',
               'subject'
@@ -360,7 +349,7 @@ describe('conventional-commits-parser', () => {
           it('should not be parsed if missing type', () => {
             const commit = parser.parse(': subject')
 
-            util(commit).negativeUtil([
+            negativeUtil(commit, [
               'type',
               'scope',
               'subject'
@@ -370,7 +359,7 @@ describe('conventional-commits-parser', () => {
           it('should not be parsed if missing subject', () => {
             const commit = parser.parse(':type:(scope):')
 
-            util(commit).negativeUtil([
+            negativeUtil(commit, [
               'type',
               'scope',
               'subject'
--- a/packages/conventional-commits-parser/src/CommitParser.spec.ts
+++ b/packages/conventional-commits-parser/src/CommitParser.spec.ts
@@ -296,9 +296,9 @@ describe('conventional-commits-parser', () => {
           }
         }
 
-        describe('positive', () => {
+        describe('positive test', () => {
 
-          it('normal usecase', () => {
+          it('should be parsed for normal use cases', () => {
             const commit = parser.parse(':type:(scope):subject')
 
             util(commit).positiveUtil([
@@ -308,16 +308,16 @@ describe('conventional-commits-parser', () => {
             ])
           })
 
-          it('missing scope', () => {
+          it('should be parsed even missing scope', () => {
             const commit = parser.parse(':type::subject')
 
             util(commit).positiveUtil(['type', 'subject'])
           })
         })
 
-        describe('negative', () => {
+        describe('negative test', () => {
 
-          it('missing initial colon', () => {
+          it('should not be parsed if the first colon is missing', () => {
             const commit = parser.parse('type:(scope):subject')
 
             util(commit).negativeUtil([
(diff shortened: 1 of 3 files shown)


packages/conventional-commits-parser/src/CommitParser.ts
```diff
@@ -204,7 +204,7 @@ export class CommitParser {
 
     if (matches) {
       correspondence.forEach((key, index) => {
-        commit[key] = matches![index + 1] || null
+        commit[key] = (matches?.groups?.[key] ?? matches[index + 1]) || null
```
dangreen, line 207:
```suggestion
        commit[key] = (matches!.groups?.[key] ?? matches![index + 1]) || null
```

packages/conventional-commits-parser/src/CommitParser.spec.ts
```diff
@@ -266,6 +266,125 @@ describe('conventional-commits-parser', () => {
...
+          it('normal usecase', () => {
+            const commit = parser.parse(':type:(scope):subject')
+
+            util(commit).positiveUtil([
+              'type',
+              'scope',
+              'subject'
+            ])
+          })
+          it('missing scope', () => {
```
dangreen, line 309:
```suggestion
          })

          it('missing scope', () => {
```

packages/conventional-commits-parser/src/CommitParser.spec.ts
```diff
@@ -266,6 +266,125 @@ describe('conventional-commits-parser', () => {
...
+            const commit = parser.parse(':type::subject')
+
+            util(commit).positiveUtil(['type', 'subject'])
+          })
+        })
+        describe('negative', () => {
```
dangreen, line 315:
```suggestion
        })

        describe('negative', () => {
```

dangreen: and etc

packages/conventional-commits-parser/src/CommitParser.spec.ts
```diff
@@ -266,6 +266,125 @@ describe('conventional-commits-parser', () => {
...
+
+        function util(commit) {
+          return {
+            positiveUtil: tokens => positiveUtil(commit, tokens),
+            negativeUtil: tokens => negativeUtil(commit, tokens)
+          }
+        }
+
+        describe('positive', () => {
+          it('normal usecase', () => {
```
dangreen, line 300: Please use names like

```
it('should ...
```

packages/conventional-commits-parser/src/CommitParser.spec.ts
```diff
@@ -266,6 +266,125 @@ describe('conventional-commits-parser', () => {
...
+            const commit = parser.parse(':type:(scope):subject')
+
+            util(commit).positiveUtil([
+              'type',
+              'scope',
+              'subject'
+            ])
```
dangreen, line 307:
```suggestion
            expect(commit).toMatchObject({
              type: 'type',
              scope: 'scope',
              subject: 'subject'
            })
```

and etc

dangreen (changes requested):

packages/conventional-commits-parser/src/CommitParser.ts
misogihagi: In my opinion, other requests can be a matter of  taste, but I think this is different.
The non-null assertion operator asserts that the operand is non-null and non-undefined, `groups` is undefined when there is no named capture group.
I think non null assertion operator should not be used.

dangreen: Ok. But you should use it in right hand

dangreen:
```suggestion
        commit[key] = (matches?.groups?.[key] ?? matches![index + 1]) || null
```

dangreen: Stop. Why `matches` here can be null or undefined if we check this value on line 205 ?

dangreen: We need here assertion because we use this value in fn callback - ts can't guarantee that here matches is not empty, but we known, because this fn will be called immediately under this condition

misogihagi: > Stop. why matches here can be null or undefined if we check this value on line 205 ?

You are right, matches is not null, matches was RegExpMatchArray or null before line 205. But after line 205, matches is only RegExpMatchArray.

> We need here assertion because we use this value in fn callback - ts can't guarantee that here matches is not empty, but we known, because this fn will be called immediately under this condition

callback is irrelevant.
What you said is true. typescript compiler doesn't know if `matches.group` is empty or not. This can't be known without executing it.
Depending on the given string and regular expression, `matches.group` may or may not be empty.
So I think it's inappropriate to use a non-null assertion here. Optional chaining is suitable.

Here is an example:
```
'text'.match(/(?<test>text)/).groups  // Object { test: "text" }
'text'.match(/text/).groups // undefined
```

In addition, [TypeScript Handbook](www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html) said `?.` is designed for optional property.

> At its core, optional chaining lets us write code where TypeScript can immediately stop running some expressions if we run into a null or undefined. The star of the show in optional chaining is the new ?. operator for optional property accesses. When we write code like

dangreen: @misogihagi In my original code suggestion i didn't touched `groups?.[key]` - it's ok.

My suggestion was about `match?.` + `match.` -> `match!.`

dangreen: we don't need `?.` here because we checked `matches` value on line 205, `matches` can't be empty. But we have `error TS18047: 'matches' is possibly 'null'.` ts error (because it inside a callback function). To ignore that we should use `!`

misogihagi: Yes, we checked `matches` on line 205. So, after 205, TS compiler knows `matches` is not null.
but `groups` property is optional property.
So I suggest to use optional chaining to confirm whether undefined or not.

misogihagi: Okay, let's go through it step by step.

1. you don't need an assertion to tell the compiler what it knows
The if statement checked to see if the matches were null.
So under those conditions, matches is a RegExpMatchArray.

    image

    If I access a property that doesn't exist, I get an error, so the compiler knows what property it is.

    image

    This is the same in the callback function
    You don't need an assertion to bother letting the compiler know what it knows.


    So the only way to retrieve the group property is to do matches.group and matches? But.


2. optional properties are safe to use with optional chains, because it's easier to know that they are optional properties

    * matches.group
    * matches?.group
    The difference between the two does not directly affect the running, but I recommend using an optional chain. That it is safer.

    The compiler will know that the `group` property is optional, but this writing improves maintainability. Tt tells us that `group` may be undefined.
    `matches.group` feels like there is always present.

dangreen: > you don't need an assertion to tell the compiler what it knows

https://github.com/conventional-changelog/conventional-changelog/actions/runs/14806428346/job/41593175331?pr=1372

misogihagi: Sorry.
This is my fault. I was using the newer Node.js and pnpm.

But I figured out our wires crossed!

github workflows shows using Node.js 18 and pnpm 8, package.json shows using typescript `5.2.2`.
My environment adopted typescript latest so compiler knows `matches` as not null.

I've investigated and found that the changes here have made the typescript compiler smarter.

- 5.2.2
- 5.3.3
- 5.4.2
Fixing the typescript compiler to each version, 
`pnpm install && pnpm test:types`, the compiler said the following:

5.2.2 and 5.3.3:
```
packages/git-client test:types: ../conventional-commits-parser/src/CommitParser.ts(207,50): error TS18047: 'matches' is possibly 'null'.
packages/git-client test:types: Failed
```

5.4.2:
```
packages/conventional-commits-parser test:types: src/CommitParser.spec.ts(280,31): error TS7006: Parameter 'commit' implicitly has an 'any' type.
packages/conventional-commits-parser test:types: src/CommitParser.spec.ts(280,39): error TS7006: Parameter 'tokens' implicitly has an 'any' type.
packages/conventional-commits-parser test:types: src/CommitParser.spec.ts(281,27): error TS7006: Parameter 'token' implicitly has an 'any' type.
packages/conventional-commits-parser test:types: src/CommitParser.spec.ts(286,31): error TS7006: Parameter 'commit' implicitly has an 'any' type.
packages/conventional-commits-parser test:types: src/CommitParser.spec.ts(286,39): error TS7006: Parameter 'tokens' implicitly has an 'any' type.
packages/conventional-commits-parser test:types: src/CommitParser.spec.ts(287,27): error TS7006: Parameter 'token' implicitly has an 'any' type.
packages/conventional-commits-parser test:types: src/CommitParser.spec.ts(288,49): error TS2554: Expected 0 arguments, but got 1.
packages/conventional-commits-parser test:types: src/CommitParser.spec.ts(292,23): error TS7006: Parameter 'commit' implicitly has an 'any' type.
packages/conventional-commits-parser test:types: src/CommitParser.spec.ts(294,27): error TS7006: Parameter 'tokens' implicitly has an 'any' type.
packages/conventional-commits-parser test:types: src/CommitParser.spec.ts(295,27): error TS7006: Parameter 'tokens' implicitly has an 'any' type.
packages/conventional-commits-parser test:types: Failed
```

I found the following statement in the release notes: assertions are no longer required starting with 5.4.

> Here, TypeScript decided that it wasn’t “safe” to assume that url was actually a URL object in our callback function because it was mutated elsewhere; however, in this instance, that arrow function is always created after that assignment to url, and it’s also the last assignment to url.
> 
> TypeScript 5.4 takes advantage of this to make narrowing a little smarter. When parameters and let variables are used in non-hoisted functions, the type-checker will look for a last assignment point. If one is found, TypeScript can safely narrow from outside the containing function. What that means is the above example just works now.
> 
> Note that narrowing analysis doesn’t kick in if the variable is assigned anywhere in a nested function. This is because there’s no way to know for sure whether the function will be called later.

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-4.html#preserved-narrowing-in-closures-following-last-assignments

It turns out that with the new compiler, I don't need to make any assertions.

PR #1373, opened 2025-05-16
fix(conventional-commits-parser): respect comment char while truncate to scissors

omertuc:
This commit fixes the conventional-commits-parser to correctly handle the scissors line when a custom comment character is used.

There was correct handling of the commentChar in general, and correct handling of the scissor line, but their combination was broken as the scissor line had the default comment character (#) hardcoded.


packages/conventional-commits-parser/src/CommitParser.ts
```diff
@@ -403,9 +403,11 @@ export class CommitParser {
       throw new TypeError('Expected a raw commit')
     }
 
-    const commentFilter = getCommentFilter(this.options.commentChar)
+    let commentChar = this.options.commentChar || '#';
```
dangreen, line 406: add default value to `src/options.ts`

packages/conventional-commits-parser/src/CommitParser.ts
```diff
@@ -403,9 +403,11 @@ export class CommitParser {
...
+
+    const commentFilter = getCommentFilter(commentChar)
     const rawLines = trimNewLines(input).split(/\r?\n/)
-    const lines = truncateToScissor(rawLines).filter(line => commentFilter(line) && gpgFilter(line))
+    const lines = truncateToScissor(rawLines, `${commentChar} ------------------------ >8 ------------------------`).filter(line => commentFilter(line) && gpgFilter(line))
```
dangreen, line 410: just pass only comment char

```suggestion
    const lines = truncateToScissor(rawLines, commentChar).filter(line => commentFilter(line) && gpgFilter(line))
```

packages/conventional-commits-parser/src/utils.ts
```diff
@@ -50,8 +50,8 @@ export function getCommentFilter(char: string | undefined) {
  * @param lines
  * @returns Lines before the scissor.
  */
-export function truncateToScissor(lines: string[]) {
-  const scissorIndex = lines.indexOf(SCISSOR)
+export function truncateToScissor(lines: string[], scissor: string) {
+  const scissorIndex = lines.indexOf(scissor)
```
dangreen, line 54:
```suggestion
  const scissorIndex = lines.indexOf(`${commentChar} ${SCISSOR}`)
```

dangreen (changes requested):

dangreen (approved):

dangreen:
@omertuc thanks!

PR #1396, opened 2025-08-04
refactor: replace `fd-package-json` with `empathic`

SuperchupuDev:
Remember #1289? It turns out that since then, a more definitive package has emerged for these types of usages, and it has [zero](https://npmgraph.js.org/?q=empathic) subdependencies instead of [one](https://npmgraph.js.org/?q=fd-package-json). Small difference but it's one less! :tada:

### DIFF
--- a/packages/conventional-changelog/package.json
+++ b/packages/conventional-changelog/package.json
@@ -69,7 +69,7 @@
     "conventional-changelog-preset-loader": "workspace:^",
     "conventional-changelog-writer": "workspace:^",
     "conventional-commits-parser": "workspace:^",
-    "fd-package-json": "^2.0.0",
+    "empathic": "^2.0.0",
     "meow": "^13.0.0",
     "normalize-package-data": "^7.0.0"
   },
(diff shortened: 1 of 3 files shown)


SuperchupuDev:
apparently not much, i've just talked to the `fd-package-json` author (who nowadays recommends `empathic`) and `fd-package-json` isn't going anywhere and will keep being maintained, so either can be used really. it'd benefit users who already have `empathic` in their tree. so your choice on what to use :+1:. although `empathic`'s bundle size is 3x smaller than `fd-package-json` but it's not a huge deal as both of them are under 1kb

`empathic`: empathic bundle size
`fd-package-json`: empathic bundle size

dangreen:
@ghiscoding Hi. Do you have plans to replace fd-package-json with empathic in lerna-lite ?

dangreen:
> it'd benefit users who already have empathic in their tree

fd-package-json have more weekly downloads, so more possible that user have fd-package-json in deps tree than empathic

> bundle size is 3x smaller

It doesn't matter for nodejs libs, to my mind here more important is install size, which is less for fd-package-json

@SuperchupuDev Sorry, I prefer to stay with fd-package-json

Anyway, thank you for you work !

PR #1420, opened 2025-12-26
feat(conventional-commits-parser): set breaking as true when the parser find it is a breaking change commit

hushicn:
Add a breaking member variable to `CommitBase` interface and set it as true in `parseBreakingHeader()` when it is pushing a "BREAKING CHANGE" to `notes`. Then the default release rule `{ breaking: true, release: "major" }` defined in @semantic-release/commit-analyzer can take effect.


dangreen (changes requested):
Please add tests

PR #1425, opened 2026-01-29
feat(conventional-changelog-conventionalcommits): filter duplicate commit references

mlmoravek:
Currently, when I generate a changelog that contains duplicate issue references in the commit, the issue is referenced multiple times: 
image

I added a filter to the references to stop the same element appearing twice in the array.


dangreen (changes requested):
Hi. Could you please add test caes?

Issue #99, opened 2015-09-02
Issue prefixes are replaced with #


rsids:
When specifying multiple issue prefixes, all the issues in the changelog are prefixed with the same prefix (#). I'd expect that the prefix would be preserved.

stevemao:
hm.. Currently only `reference.raw` exposes the prefix. We should add a field `referencePrefix`
```
expect(msg.references).to.eql([{
    action: 'handled',
    owner: 'angular',
    repository: 'angular.js',
    issue: '1',
    raw: 'angular/angular.js#1',

    // referencePrefix: '#'
}]);
```

stevemao:
tracking at https://github.com/stevemao/conventional-commits-parser/issues/17

Issue #1461, opened 2026-05-29
Changelog renders "closes" for bare references, causing unintended GitHub issue auto-closing


afreyermuth98:
### Description

The `conventionalcommits` preset renders the literal keyword **`closes`** in front of *every* issue reference found in a commit message —
including bare mentions that carry **no closing action keyword**.

The parser already distinguishes these: a reference like `Closes #1` gets `action: "Closes"`, while a plain mention (`part of #123`, `Refs:
#100`, `see #1`) gets `action: null`. The writer ignores that distinction and labels them all `closes`.

This is incorrect on its own, and it has a concrete side effect. A common release-automation pattern is to reuse the generated changelog as a
**pull-request body** (release PRs). GitHub honors closing keywords anywhere in a PR body, so a changelog line like `, closes #123` makes
**merging the release PR auto-close issue #123** — even though no commit ever intended to close it. The author only *mentioned* it.

This is the same class of problem as #1288 ("the link is printed after the 'closes' keyword … even though the commit doesn't actually close the
comment").

### Steps to reproduce

Commit:

```
feat: some feature

Part of #123
```

Generated changelog:

```
* some feature (abc1234), closes #123
```

### Expected

`#123` is a bare mention (`action: null`), so it should **not** be rendered as `closes`. Only references with a closing action keyword
(`closes`/`fixes`/`resolves`/…) should appear under `closes`.

### Actual

`#123` is rendered as `, closes #123`, which GitHub treats as a directive to auto-close issue #123 when the changelog is used in a PR body.

### Proposed fix

Split references by their parsed `action`:
- references with a closing action → rendered under `, closes`
- all other references (bare mentions) → rendered under `, refs`

Links are preserved in both cases; only the keyword changes. PR to follow.

suzunn:
The safest acceptance criterion here is to pin the behavior at the writer boundary, not only in the final changelog string. I would keep the generated text assertion, but also cover the transformed reference grouping for three cases: a normal closing action (`Closes #1`), a bare mention (`part of #2`), and the ambiguous token-footer form (`Fixes: #3` / `Refs: #4`) that currently arrives from the parser with `action === null`.

That gives the maintainers a clear contract: the writer only prints `closes` when the parser explicitly marks the reference as a closing action. If the parser contract changes later, the test failure will point to the intentional decision point instead of letting release PR bodies regain accidental auto-closing behavior.

PR #1462, opened 2026-05-29
fix: only render "closes" for closing-action references

afreyermuth98:
REF https://github.com/conventional-changelog/conventional-changelog/issues/1461

### DIFF
--- a/packages/conventional-changelog-conventionalcommits/src/templates.js
+++ b/packages/conventional-changelog-conventionalcommits/src/templates.js
@@ -47,22 +47,28 @@ export const commitPartial = `*{{#if scope}} **{{scope}}:**
 {{~/if}}{{~/if}}
 
 {{~!-- commit references --}}
-{{~#if references~}}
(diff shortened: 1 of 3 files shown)


afreyermuth98:
cc @dangreen @llamington

suzunn:
I like the direction here because it keeps the parser's `reference.action` signal intact instead of trying to infer intent from the rendered text later. One thing I would tighten before merge is the token-footer case in the regression test: `Fixes: #99` reads like a closing keyword to users, but the parser currently emits it as `action === null`, so this change deliberately moves it to `refs`.

That behavior may be correct for the current parser contract, but it is subtle enough that I would add a short fixture/assertion that checks both the rendered output and the transformed `closingReferences` / `otherReferences` split for that input. It would make future parser changes safer: if the parser starts treating token-style footers as closing actions, the expected changelog behavior can be updated intentionally instead of silently changing release PR auto-close semantics.

PR #1466, opened 2026-06-16
fix(‎conventional-changelog): improve version tag parsing in changelog generation

### LINKED ISSUE #1465: v7 omits the latest release from the changelog when its tag is on HEAD
Starting from v7, when the latest tag points at HEAD and the branch has been pushed, `conventional-changelog -r 0` omits the section for the latest release. The previous version (`conventional-changelog-cli@5` before the merge) generates the section correctly for the same repo.

**Steps to reproduce:**

```sh
mkdir repro && cd repro && git init -q -b main
npm init -y >/dev/null
npm i -D conventional-changelog@7.2.0 conventional-changelog-angular
git add package.json && git commit -qm "feat: first" && git tag v1.0.0
git commit -qm "feat: second" --allow-empty
npm pkg set version=1.1.0
git add package.json && git commit -qm "chore: release v1.1.0" && git tag v1.1.0
git init -q --bare ../origin.git && git remote add origin ../origin.git
git push -q origin main --tags

npx conventional-changelog -p angular -r 0 --stdout
```

**Expected:** sections for both `1.1.0` (containing `feat: second`) and `1.0.0`. This is the output of `conventional-changelog-cli@5 -p angular -r 0` for the same repo.

**Actual:** only the `1.0.0` section. The `1.1.0` section is missing. With `-u`, the missing commits appear under an `Unreleased` heading with an incorrect compare link (`v1.1.0...<commit hash>`), even though they belong to the tagged v1.1.0.

The `git push` is required to reproduce. The trigger is another ref appearing after the tag in the commit decoration; remote-tracking refs (`origin/main`, `origin/HEAD`) sort after tags in `%d` output, which is the format used to fetch commits. After the push, `gitTags` for the tagged HEAD commit is `(HEAD -> main, tag: v1.1.0, origin/main, origin/HEAD)`.

The version regexes in `packages/conventional-changelog/src/utils.ts` match greedily:

```js
export const versionTagRegex = /tag:\s*(.*)[,)]/i
export const defaultVersionRegex = /tag:\s*[v=]?(.*)[,)]/i
```

Against the decoration above, `(.*)` captures `1.1.0, origin/main, origin/HEAD` instead of `1.1.0`. That is not a valid version, so the writer never recognizes the commit as the start of the v1.1.0 release. With no release boundary, the commits since v1.0.0 are treated as unreleased instead of as part of v1.1.0. Unreleased commits are dropped by default when the latest tag matches the package.json version (the version is considered already released). Tags not on HEAD are unaffected because their decoration is just `(tag: v1.0.0)`.

The equivalent regexes in v6 were non-greedy, e.g. in `conventional-changelog-core/lib/merge-config.js`:

```js
/tag:\s*[v=]?(.+?)[,)]/gi
```

The greedy variants date from the v7 rewrite (42556a1, #1352). Restoring `(.+?)` in the three regexes in `utils.ts` (the two constants above and the `tags.prefix` variant in `defaultCommitTransform`) resolves the issue.

Tested on 7.2.0 (Node 22, macOS). The regexes are unchanged since v7.0.0, so 7.0.0–7.2.0 should all be affected.

peiyanlu:
Fixes #1465

### Problem

Changelog generation may produce incorrect or missing version titles when parsing git commit metadata containing release tags.

This issue occurs when `origin/<branch>` points to a commit that includes a version tag, causing inconsistent parsing results.

---

### Root cause

Current regex patterns used for extracting tags:

```ts id="9c1k2a"
/tag:\s*(.*)[,)]/i
/tag:\s*[v=]?(.*)[,)]/i
```

can incorrectly match additional git ref metadata (e.g. `origin/<branch>`, `HEAD`), leading to unreliable tag extraction.

---

### Changes

* Refined regex to more strictly extract version tag values
* Avoid matching unrelated git ref decorations
* Make tag parsing independent of `HEAD` / branch state

---

### Verification

Tested against multiple repository states:

* Cases where `origin/<branch>` points to feature commits → OK
* Cases where `origin/<branch>` points to release commits → fixed incorrect parsing
* Multiple release history scenarios → consistent output

---

### Impact

* More reliable changelog version detection
* Fixes missing/incorrect title in some release scenarios

### DIFF
--- a/packages/conventional-changelog/src/ConventionalChangelog.spec.ts
+++ b/packages/conventional-changelog/src/ConventionalChangelog.spec.ts
@@ -24,6 +24,7 @@ const {
   tearsWithJoy
 } = BetterThanBefore()
 let testTools: TestTools
+const BARE_REPO = '../bare-origin.git'
 
 setups([
   () => { // 1
@@ -144,12 +145,13 @@ setups([
     testTools.exec('git merge feature2 -m"Merge branch \'feature2\'"')
   },
   () => { // 20
+    testTools?.cleanup()
     testTools = new TestTools()
     // mock remote
-    testTools.exec('git init --bare ../origin20.git')
+    testTools.exec(`git init --bare ${BARE_REPO}`)
 
     testTools.gitInit()
-    testTools.exec('git remote add origin ../origin20.git')
+    testTools.exec(`git remote add origin ${BARE_REPO}`)
 
     testTools.writeFileSync('package.json', JSON.stringify({
       name: 'conventional-changelog',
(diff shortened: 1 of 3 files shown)


packages/conventional-changelog/src/index.ts
```diff
@@ -1,3 +1,4 @@
 export type * from './types.js'
 export * from './ConventionalChangelog.js'
 export * from './cli/cli.js'
+export { guessNextTag, isUnreleasedVersion, defaultCommitTransform } from './utils.js'
```
dangreen: why?

dangreen (changes requested):
Please add tests

packages/conventional-changelog/src/index.ts
peiyanlu: > why?

I need to export these utilities because users often override `transformCommit`, but in most cases they don’t want to fully replace the default behavior — they want to extend it.

packages/conventional-changelog/src/ConventionalChangelog.spec.ts
```diff
@@ -136,6 +136,112 @@ setups([
...
+    testTools.gitCommit('feat: third commit')
+
+    // v1.0.3
+    testTools.writeFileSync('./package.json', '{"version": "1.0.3"}')
+    testTools.exec('git add --all')
+    testTools.gitCommit('chore(release): v1.0.3')
+    testTools.exec('git tag v1.0.3')
+
+    // push
+    testTools.exec('git push -f origin master --tags')
```
dangreen: Pretty sure that this is AI generated code. 

I am not against using AI, but please control and review that you push here.

There no any `git push` in any tests in this repo. You should remove it.

Secondly, You don't need many tags to test your code. 2-3 tags is enough

dangreen (changes requested):

packages/conventional-changelog/src/ConventionalChangelog.spec.ts
peiyanlu: Thanks for the feedback.

Regarding `git push`, I'm not sure whether my analysis in #1465 is correct, so I'd appreciate your opinion on it.

I also reproduced the issue in a local test repository ([`config-test`](https://github.com/peiyanlu/config-test)) and found that the problem doesn't occur unless the commits are pushed. That's why I simulated a local `git push` in the test.

If `git push` should not be used in the test suite, could you suggest a better way to verify this behavior? I don't have much experience writing this kind of Git-related test, so I'd really appreciate any guidance on how you would approach it.

I can simplify the test cases to use only 2–3 tags while still covering the intended behavior.

I'll update the PR accordingly. Thanks for the review.

dangreen: `git push` can't affect changelog generation

peiyanlu: I understand now — my original assumption was incorrect. I've updated the implementation accordingly. Thanks for pointing that out.

dangreen (approved):

PR #1484, opened 2026-06-26
feat(conventional-changelog-conventionalcommits)!: support commit type effects

### LINKED ISSUE #1476: feat(conventionalcommits): add `noBump` type property to decouple bump behavior from changelog visibility
## Problem

When using `conventional-changelog-conventionalcommits` (in our case via `@release-it/conventional-changelog`), I expected that only commit types with explicit semver meaning: `feat` (minor), `fix` (patch) and any type with `!` (major), would trigger a version bump, per the [Conventional Commits spec](https://www.conventionalcommits.org/).

**What actually happens:** every commit type triggers a `patch` bump, including `docs`, `chore`, `ci`, `test`, etc.

After investigation, this appears to be by design in `whatBump.js`:

```javascript
// whatBump.js , createWhatBump()
return function whatBump(commits) {
  let level = 2  // ← default: patch for ANY commit
  // ...
  commits.forEach((commit) => {
    // only feat/fix/BREAKING actually change `level`
    // but everything else still returns level=2 (patch)
  })
}
```

So any recognized commit, regardless of type, produces a `patch` bump by default.

This is not a regression, it's the **original design**. The `let level = 2` has been present since the file was created. Looking at the commit history of `whatBump.js`:

- [`90abd8a`](https://github.com/conventional-changelog/conventional-changelog/commit/90abd8a8a18d7aab06adcb67b36d7663e663576c) (Feb 24, 2025) — restructured the file; `let level = 2` already present
- [`e2973e2`](https://github.com/conventional-changelog/conventional-changelog/commit/e2973e28380405bfa62396ad0638ffc7cc237e0d) (May 18, 2025, @dangreen) — added `bumpStrict` as an opt-in fix, which **implicitly acknowledges the problem** but doesn't change the default for backward compatibility

The `bumpStrict` addition shows the maintainers are aware of this behavior and chose to fix it as opt-in rather than change the default (which would be a breaking change). Fair enough, but the current `bumpStrict` implementation relies on `hidden` to determine which types to exclude, which creates a coupling problem: you can't exclude a type from bumping without also hiding it from the changelog. 

### Option 1: `hidden: true`

Setting `hidden: true` on non-releasing types removes them from the changelog entirely. This doesn't work for our use case, we want `docs:`, `chore:`, `ci:` commits **visible in the changelog** for auditability, just without triggering a release.

### Option 2: `bumpStrict: true`

`bumpStrict` was added (thank you @dangreen!) to address the unwanted bump issue. When enabled, it builds a list of types to ignore from the bump calculation based on `type.hidden`:

```javascript
const hiddenTypes = bumpStrict && types.reduce((hiddenTypes, type) => {
  if (type.hidden) {          // ← only checks `hidden`
    hiddenTypes.push(type.type)
  }
  return hiddenTypes
}, [])
```

The problem: `bumpStrict` relies on `hidden` to determine which types should not bump. So to prevent `docs:` from bumping, you must set `hidden: true` on it, which also removes it from the changelog.

**`bumpStrict: true` + `hidden: true` = no bump AND no changelog entry.**

There is no combination of existing options that gives: **visible in changelog + no version bump**.

## Proposal: `noBump` property

Add a per-type `noBump` boolean property that controls bump behavior independently of changelog visibility.

### The change

**One line** in `whatBump.js`:

```diff
   const hiddenTypes = bumpStrict && types.reduce((hiddenTypes, type) => {
-    if (type.hidden) {
+    if (type.hidden || type.noBump) {
       hiddenTypes.push(type.type)
     }
     return hiddenTypes
   }, [])
```

### Behavior matrix

| `hidden` | `noBump` | Changelog | Bump |
|----------|----------|-----------|------|
| `false`  | `false`  | ✅ Visible | ✅ Bump |
| `false`  | `true`   | ✅ Visible | ❌ No bump |
| `true`   | `false`  | ❌ Hidden  | ✅ Bump |
| `true`   | `true`   | ❌ Hidden  | ❌ No bump |

### Example usage

```json
{
  "types": [
    { "type": "feat", "section": "Features" },
    { "type": "fix", "section": "Bug Fixes" },
    { "type": "docs", "section": "Documentation", "noBump": true },
    { "type": "chore", "section": "Chores", "noBump": true },
    { "type": "ci", "section": "CI", "noBump": true },
    { "type": "test", "section": "Tests", "noBump": true }
  ],
  "bumpStrict": true
}
```

`docs:` and `chore:` commits appear in the changelog but do **not** trigger a version bump.

### Properties of this change

- **Non-breaking:** `noBump` defaults to `undefined`/`false`, existing configs work identically
- **Minimal:** 1 line of logic in `whatBump.js`
- **Only takes effect with `bumpStrict: true`** , same gate as the existing `hidden`-based filtering
- **No impact on changelog generation**, `writer.js` still uses `hidden` exclusively for visibility

## I'm happy to submit a PR

I've traced the code path through `whatBump.js`, `writer.js`, and `constants.js`. I can submit a minimal PR with:
- The one-line fix in `whatBump.js`
- Tests covering the new property
- README update documenting `noBump`

Let me know if this approach works or if you'd prefer a different design (e.g. different property name, different semantics).

dangreen:
BREAKING CHANGE: the commit type `effect` property replaces the `hidden` commit type property and the `bumpStrict` preset option for controlling changelog visibility and version bumps.

Fixes #1476.

### DIFF
--- a/packages/conventional-changelog-conventionalcommits/README.md
+++ b/packages/conventional-changelog-conventionalcommits/README.md
@@ -83,12 +83,21 @@ See [conventional-changelog-config-spec](https://github.com/conventional-changel
 |--------|-------------|
 | ignoreCommits | Regular expression to match and exclude raw commits before parsing. |
 | issuePrefixes | Array of issue prefixes parsed as issue references. Defaults to `['#']`. |
-| types | Array of commit type objects defining sections, hidden types, and optional scope-specific type handling. Default value is available via the `DEFAULT_COMMIT_TYPES` export. |
+| types | Array of commit type objects defining sections, release effects, and optional scope-specific type handling. Default value is available via the `DEFAULT_COMMIT_TYPES` export. |
 | scope | String or array of scope names to filter commits. Only commits with matching scopes will be included. When `scopeOnly` is `false` (default), commits without any scope are also included. |
 | scopeOnly | When `true` and `scope` is specified, excludes commits that have no scope. When `false` (default), includes both scoped and unscooped commits when filtering by scope. |
 | preMajor | When `true`, breaking changes and features are downgraded by one semver level before the first major release. |
-| bumpStrict | When `true`, version bumps occur only for breaking changes or non-hidden commit types. When `false` (default), any commit can trigger a version bump. |
 | formatIssueUrl | Function that formats issue reference URLs. Receives `(context, reference)`. |
 | formatCommitUrl | Function that formats commit URLs. Receives `(context, commit)`. |
 | formatCompareUrl | Function that formats release comparison URLs. Receives `(context)`. |
 | formatUserUrl | Function that formats user mention URLs. Receives `(context, user)`. |
+
+Each item in `types` can define an `effect`:
+
+| Effect | Changelog | Version bump |
+|--------|-----------|--------------|
+| `bump` | Included | Included |
+| `changelog` | Included | Ignored |
+| `hidden` | Hidden | Ignored |
+
+`effect` defaults to `bump` when omitted. Breaking changes always trigger a major bump and are included in the changelog regardless of the type effect.
--- a/packages/conventional-changelog-conventionalcommits/src/constants.js
+++ b/packages/conventional-changelog-conventionalcommits/src/constants.js
@@ -1,57 +1,62 @@
 export const DEFAULT_COMMIT_TYPES = Object.freeze([
   {
     type: 'feat',
-    section: 'Features'
+    section: 'Features',
+    effect: 'bump'
   },
   {
     type: 'feature',
-    section: 'Features'
+    section: 'Features',
+    effect: 'bump'
   },
   {
     type: 'fix',
-    section: 'Bug Fixes'
+    section: 'Bug Fixes',
+    effect: 'bump'
   },
   {
     type: 'perf',
-    section: 'Performance Improvements'
+    section: 'Performance Improvements',
+    effect: 'bump'
   },
   {
     type: 'revert',
-    section: 'Reverts'
+    section: 'Reverts',
+    effect: 'bump'
   },
   {
     type: 'docs',
     section: 'Documentation',
-    hidden: true
+    effect: 'hidden'
   },
   {
     type: 'style',
     section: 'Styles',
-    hidden: true
+    effect: 'hidden'
   },
   {
     type: 'chore',
     section: 'Miscellaneous Chores',
-    hidden: true
+    effect: 'hidden'
   },
   {
     type: 'refactor',
     section: 'Code Refactoring',
-    hidden: true
+    effect: 'hidden'
   },
   {
     type: 'test',
     section: 'Tests',
-    hidden: true
+    effect: 'hidden'
   },
   {
     type: 'build',
     section: 'Build System',
-    hidden: true
+    effect: 'hidden'
   },
   {
     type: 'ci',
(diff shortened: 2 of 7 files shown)


LeCrew163:
Thanks for this change!

Funny enough, we need the exact opposite of it. 🙂

Is there a way to get the pre-v10 behaviour back — any recognised commit bumping at least a patch — without adding a types list to every repo? We version a few dozen infrastructure repos where the history is mostly chore(deps): from Renovate, and there a dependency bump really is the thing we ship. They quietly stopped releasing after v10.

We looked at the angular preset, since its whatBump still starts at let level = 2. But it ignores the preset config entirely (createPreset doesn't pass it to createParserOpts/createWriterOpts), so we'd lose issuePrefixes and issueUrlFormat — we link commits to our internal Jira, so that's not an option for us.

Is there a preset- or CLI-level switch we're missing?

dangreen:
Thanks for the detailed context!

Short answer: no, there's no preset- or CLI-level switch for this, and that's intentional. After #1484 the bump behaviour is driven by a single mechanism — the `effect` property on commit types — and we'd like to keep it that way rather than reintroduce a second global toggle. You're also right that the `angular` preset doesn't forward preset config to its parser/writer options — it's intentionally kept minimal and non-configurable; `conventionalcommits` is the configurable one.

That said, I think your case has a cleaner solution than restoring "everything bumps":

**1. Let Renovate emit `fix(deps)` instead of `chore(deps)` (recommended).**
With semantic commits, Renovate already uses `fix(deps):` for `dependencies` and `chore(deps):` for `devDependencies` — which matches your situation exactly: if a dependency bump is the thing you ship, it *is* a `fix`. For managers where depType classification is off (Docker, Terraform, Helm are common offenders in infra repos), you can force it centrally in your shared Renovate config:

```json
{
  "packageRules": [
    { "matchManagers": ["dockerfile", "terraform", "helm-values"], "semanticCommitType": "fix" }
  ]
}
```

Since it lives in the shared preset, it's one change for all your repos. Side benefit: `chore` was `hidden` even before v10, so your releases had empty changelogs — with `fix(deps)` the bumps will actually show up in the notes, linked to your Jira via `issuePrefixes`/`issueUrlFormat` as before.

**2. A tiny shared preset, if you'd rather not touch commit messages.**
You don't need to maintain a full `types` list per repo — `DEFAULT_COMMIT_TYPES` is exported, and type entries match on `type` + `scope`, so it's enough to prepend one entry:

```js
// @your-org/changelog-preset
import createPreset, { DEFAULT_COMMIT_TYPES } from 'conventional-changelog-conventionalcommits'

export default config => createPreset({
  types: [
    { type: 'chore', scope: 'deps', section: 'Dependencies', effect: 'bump' },
    ...DEFAULT_COMMIT_TYPES
  ],
  ...config
})
```

(Drop `scope` if you want every `chore` to bump.) Publish it once, and each repo just points `-p @your-org/changelog-preset` — one dependency line instead of a config blob.

jdbruijn:
It took me soo long to get here and finally figure out why `hidden: true` wasn't working anymore. Shouldn't this change have been documented in a new version of https://github.com/conventional-changelog/conventional-changelog-config-spec as well? I was initially looking there and figured `hidden: true` was the correct syntax. Now after seeing the release notes here I got to this PR and changed to `effect: 'hidden'`, finally hiding the section ofter hours going around.

PR #1485, opened 2026-06-26
fix(conventional-commits-parser,template): keep footers out of breaking notes

### LINKED ISSUE #428: BREAKING CHANGE is too greedy
the BREAKING CHANGE is greedy, in that it consumes text until the end of the commit message after the `:`; this perhaps makes sense for the body? but for the footer which can contain multiple lines of meta information, we should probably only consume tokens until the end of one line of the footer message if we see a BREAKING CHANGE.

### LINKED ISSUE #773: conventional commits parser does not accurately parse commit footers
The conventional commit spec indicates the footer can be indicated by the git trailers format

```txl
footers other than BREAKING CHANGE: <description> may be provided and follow a convention similar to git trailer format.
```

I'm having trouble making this work without explicit section marking

```javascript
const parser = require('conventional-commits-parser')
const commit = `
test: this is a test

this is the body of the commit
 
Foo: This is the footer of the commit
Bar: This is a git trailer syntax
`.trim()

parser.sync(commit)
{
  type: 'test',
  scope: null,
  subject: 'this is a test',
  merge: null,
  header: 'test: this is a test',
  body: 'this is the body of the commit\n' +
    '\n' +
    'Foo: this is a trailer\n' +
    'Bar: This is the footer of the commit',
  footer: null,
  notes: [],
  references: [],
  mentions: [],
  revert: null
}
```

This should be parsed with the trailers in the footer.

```javascript
{
  type: 'test',
  scope: null,
  subject: 'this is a test',
  merge: null,
  header: 'test: this is a test',
  body: 'this is the body of the commit\n',
  footer: 'Foo: this is a trailer\nBar: This is the footer of the commit',
  notes: [],
  references: [],
  mentions: [],
  revert: null
}
```

This really only seems to work if you have issue references or breaking change notes.
You could also use explicit sections, But i wouldn't expect to have to do that in this case

### LINKED ISSUE #819: BREAKING CHANGE message cut off if it contains a reference
I made a breaking change recently and I want to add a reference back to the associated issue in the change note, however `conventional-commits-parser` will stop at the first line that contains a reference and *not* add it to the resulting message.

For example, the commit

```
feat: add some cool new stuff

BREAKING CHANGE: We removed some boring old stuff in
order to make room for the new stuff. We went back and
forth on whether or not to do this in #4, so if you want
more information you should check out that discussion.
```

Will produce the following:
```
### ⚠ BREAKING CHANGES

    *  We removed some boring old stuff in
       order to make room for the new stuff. We went back and
```

I have tracked down the issue to here:

https://github.com/conventional-changelog/conventional-changelog/blob/f1f50f56626099e92efe31d2f8c5477abd90f1b7/packages/conventional-commits-parser/lib/parser.js#L232

---

If I had to guess why that line is there I'd say it's meant to filter out trailing `Closes #x` lines, but clearly it's not a good solution for that. Preferably we'd fix #428 to make that workaround unnecessary. Perhaps only stopping the note once we reach an empty line?

dangreen:
Fixes #428.

Fixes #773.

Fixes #819.

### DIFF
--- a/packages/conventional-changelog-conventionalcommits/test/index.spec.js
+++ b/packages/conventional-changelog-conventionalcommits/test/index.spec.js
@@ -92,6 +92,13 @@ setups([
   },
   () => {
     testTools.gitCommit(['fix: replace `@nano_kit/router` with `@nano_kit/router2`'])
+  },
+  () => {
+    testTools.gitCommit([
+      'feat!: support commit type effects',
+      'BREAKING CHANGE: effect replaces hidden.',
+      'Fixes #1476.'
+    ])
   }
 ])
 
@@ -553,6 +560,20 @@ describe('conventional-changelog-conventionalcommits', () => {
 
     expect(chunks[0]).toContain('`@nano_kit/router`')
     expect(chunks[0]).not.toContain('[@nano')
+  })
+
+  it('should not include closing references in breaking change notes', async () => {
+    preparing(13)
+
+    const log = new ConventionalChangelog(testTools.cwd)
+      .readPackage()
+      .config(preset())
+      .write()
+    const chunks = await toArray(log)
+
+    expect(chunks[0]).toContain('effect replaces hidden.')
+    expect(chunks[0]).toContain(', closes [#1476](https://github.com/conventional-changelog/conventional-changelog/issues/1476)')
+    expect(chunks[0]).not.toContain('Fixes #1476')
   })
 
   describe('type effects', () => {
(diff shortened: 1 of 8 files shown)


JohnVillalovos:
FYI:  This seems to have caused a regression.

The `footerToken` regex added here (`getFooterTokenRegex`) introduces a false-positive that can break parsing of Dependabot commit messages.


For example CI is currently failing for this Dependabot update. https://github.com/LibreBooking/librebooking/pull/1604

```
chore(deps): bump foo from 1.0.0 to 2.0.0

Bumps foo from 1.0.0 to 2.0.0.

---
updated-dependencies:
- dependency-name: "foo"
  dependency-version: 2.0.0
  dependency-type: direct:development
...

Signed-off-by: dependabot[bot] <support@github.com>
```

The line `  dependency-version: 2.0.0` matches `^\s*(?:BREAKING CHANGE|[\w-]+)(?::\s+...).+`, so the parser treats it as the start of the footer.

